PR #54, opened 2020-09-15
[stable/weblate] Added chart

sietevecesmal:
## Description

Added weblate chart

## Checklist

### DIFF
--- a/stable/weblate/values.yaml
+++ b/stable/weblate/values.yaml
@@ -27,7 +27,7 @@ database:
   create: true
 
 ingress:
-  enabled: true
+  enabled: false
   annotations:
     nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
     nginx.ingress.kubernetes.io/proxy-body-size: 512m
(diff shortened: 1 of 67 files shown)


stable/weblate/values.yaml
```diff
@@ -0,0 +1,39 @@
...
+  environment:
+    WEBLATE_DEBUG: False
+    WEBLATE_LOGLEVEL: WARNING
+    WEBLATE_SITE_TITLE: Weblate
+    WEBLATE_ADMIN_NAME: "Weblate Admin"
+    WEBLATE_ENABLE_HTTPS: True
+    SECURE_SSL_REDIRECT: True
+
+database:
+  create: true
```
max-rocket-internet, line 27: This will error if the user doesn't have the postgres-controller installed so best disable by default it I think:

```suggestion
database:
  create: false
```

max-rocket-internet (approved):

max-rocket-internet:
The `install` CI test is failing because the pod crashes. This is because the app is looking for `WEBLATE_ADMIN_EMAIL` env var.

max-rocket-internet (approved):

PR #57, opened 2020-09-18
[stable/toxiproxy] Misc fixes

nreymundo:
## Description

- Fixed trying to use the same `name` for all the ports defined in `Values.service.proxies.ports`
- Fixed wrong mapping when trying to specify extra labels that should be applied to the resources. 
- Fixed wrong indentation when specifying resources. 

## Checklist


max-rocket-internet (approved):

max-rocket-internet:
You will need to bump the version number of the chart here to get the CI checks to pass: https://github.com/deliveryhero/helm-charts/blob/master/stable/toxiproxy/Chart.yaml#L4

It must be a string so keep the quotes.

PR #63, opened 2020-10-15
[stable/kube-bench] Helm chart for kube-bench

goelankitt:
## Description
This helm chart creates a cronjob for running kube-bench. At this point, it only supports eks and gke.

## Checklist


stable/kube-bench/templates/cron.yaml
```diff
@@ -0,0 +1,39 @@
...
+          volumes:
+            - name: var-lib-kubelet
+              hostPath:
+                path: "/var/lib/kubelet"
+            - name: etc-systemd
+              hostPath:
+                path: "/etc/systemd"
+            - name: etc-kubernetes
+              hostPath:
+                path: "/etc/kubernetes"
```
max-rocket-internet, line 39: It's probably nicer to include the standard options that allow the user to customise the scheduling for their env. e.g. https://github.com/deliveryhero/helm-charts/blob/master/stable/k8s-event-logger/templates/deployment.yaml#L41-L52

stable/kube-bench/values.yaml
```diff
@@ -0,0 +1,11 @@
+provider: eks
+
+cronjob:
+  schedule: "*/1 * * * *"
```
max-rocket-internet, line 4: I think every minute is too frequent?? 😅

max-rocket-internet (changes requested):
CI steps need to pass also

stable/kube-bench/values.yaml
goelankitt: yes it is. Making it monthly.

max-rocket-internet (approved):

PR #70, opened 2020-11-02
[stable/kube-bench] Use latest image of kube-bench for security tests

goelankitt:
## Description

This PR removes app version from the chart.yaml file as the chart is not tied to a specific version of kube-bench. The version of kube-bench to be used can be specified using the values.yaml file. It also sets the default value of kube-bench image tag to latest.

## Checklist

### DIFF
--- a/stable/kube-bench/values.yaml
+++ b/stable/kube-bench/values.yaml
@@ -5,7 +5,7 @@ cronjob:
 
 image:
   repository: aquasec/kube-bench
-  tag: 0.1.0
+  tag: latest
   pullPolicy: IfNotPresent
 
 resources: {}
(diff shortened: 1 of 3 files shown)


stable/kube-bench/values.yaml
```diff
@@ -5,7 +5,7 @@ cronjob:
 
 image:
   repository: aquasec/kube-bench
-  tag: 0.1.0
+  tag: latest
```
max-rocket-internet, line 8: `latest` can lead to many problems. Can we use a hash if there's no proper tag available?

max-rocket-internet (approved):
You need to bump the chart version for CI to pass

max-rocket-internet (approved):

PR #71, opened 2020-11-02
[stable/cachet] add to cachet podLabels

nyambati:
Signed-off-by: Thomas Nyambati <thomasnyambati@gmail.com>

## Checklist

### DIFF
--- a/stable/cachet/templates/deployment.yaml
+++ b/stable/cachet/templates/deployment.yaml
@@ -4,6 +4,9 @@ metadata:
   name: {{ include "cachet.fullname" . }}
   labels:
 {{ include "cachet.labels" . | indent 4 }}
+{{- if .Values.extraLabels }}
+{{ toYaml .Values.extraLabels | indent 4 }}
+{{- end }}
 spec:
   replicas: {{ .Values.replicaCount }}
   selector:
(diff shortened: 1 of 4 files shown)


max-rocket-internet (approved):

OliverKlette85 (changes requested):
Could you please also add the laabels to the deployment itself?

max-rocket-internet:
If you want to add labels to the deployment then I think you will need 2 separate values, like perhaps `podLabels` and `extraLabels` or something.

nyambati:
extraLabels will be just fine. I will make the updates.

OliverKlette85:
@max-rocket-internet why do we need an extra value? Couldn't we just rename `podLabels` to just `labels` or `extraLabels` and use it for both (deployment and pods)?

max-rocket-internet:
> Couldn't we just rename podLabels to just labels or extraLabels and use it for both (deployment and pods)

Sure 🙂

stable/cachet/templates/deployment.yaml
```diff
@@ -4,6 +4,9 @@ metadata:
   name: {{ include "cachet.fullname" . }}
   labels:
 {{ include "cachet.labels" . | indent 4 }}
+{{- if .Values.extraLabels }}
```
max-rocket-internet, line 7: You're only adding them to the deployment and pods? I think you should put it in the `_helpers.tpl` so that `extraLabels` is included in the `labels` template function, otherwise it's not very consistent?

max-rocket-internet (approved):

OliverKlette85 (approved):

PR #74, opened 2020-11-16
[stable/node-problem-detector] Update psp to explicitly allow prometheus ports

cnmcavoy:
## Description

* Updated psp to explicitly allow prometheus port range on host when hostNetwork is set

## Checklist


max-rocket-internet (approved):
Thanks @cnmcavoy

PR #83, opened 2020-11-26
[stable/locust] Add update strategy to values

zabudskyi:
## Description

Added possibility to set update strategy. Should be merged after https://github.com/deliveryhero/helm-charts/pull/82, if it looks fine.

## Checklist

### DIFF
--- a/stable/locust/Chart.yaml
+++ b/stable/locust/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: locust
-version: "0.9.2"
+version: "0.9.4"
 appVersion: 1.4.1
 home: https://github.com/locustio/locust
 icon: https://locust.io/static/img/logo.png
--- a/stable/locust/README.md
+++ b/stable/locust/README.md
@@ -91,7 +91,7 @@ helm install my-release deliveryhero/locust -f values.yaml
 | master.logLevel | string | `"INFO"` | Log level. Can be INFO or DEBUG |
 | master.resources | object | `{}` | resources for the locust master |
 | master.serviceAccountAnnotations | object | `{}` |  |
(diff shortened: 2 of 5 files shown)


zabudskyi:
I run helm-docs for Readme and output doesn't look well.
It removes `How to install this chart` section

This change doesn't look fine as `strategy` does not only have `type` but `rollingUpdate` section as option too.

```
-| master.strategy | object | `{}` |  |
+| master.strategy.type | string | `"RollingUpdate"` |  |
 | nameOverride | string | `""` |  |
 | nodeSelector | object | `{}` |  |
 | podSecurityContext | object | `{}` |  |
@@ -112,10 +86,7 @@ helm install my-release deliveryhero/locust -f values.yaml
 | worker.replicas | int | `1` |  |
 | worker.resources | object | `{}` | resources for the locust worker |
 | worker.serviceAccountAnnotations | object | `{}` |  |
-| worker.strategy | object | `{}` |  |
-
+| worker.strategy.type | string | `"RollingUpdate"` |  |
```

max-rocket-internet:
> It removes How to install this chart section

I don't see this when I checkout your branch?

> +| worker.strategy.type | string | `"RollingUpdate"` |  |

I think this makes sense, no?

max-rocket-internet (approved):
Fix the docs and then we can merge

zabudskyi:
It can be 
```
  strategy:
    type: RollingUpdate
    rollingUpdate:
      maxUnavailable: 100
      maxSurge: 50%
```
So `worker.strategy.type` doesn't look well as value description because `worker.strategy` is value to be set in this case.
I'll try to use `worker.strategy.type` to pass the check

max-rocket-internet:
Thanks @zabudskyi

PR #88, opened 2020-12-02
[stable/labelsmanager-controller]add labels manager controller

nyambati:
## Checklist

### DIFF
--- a/stable/labelsmanager-controller/values.yaml
+++ b/stable/labelsmanager-controller/values.yaml
@@ -1,0 +1,46 @@
+# Default values for labelsmanager-controller.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 1
+
+controller:
+  # controller.production -- The log level of the controller. enable for json logs
+  production: true
+  # controller.leaderElection -- Enable leader election for running multiple controller pods
+  leaderElection: false
+  # controller.metricAddr -- Address to serve prometheus metrics on. "0" is disabled.
+  metricAddr: "0"
+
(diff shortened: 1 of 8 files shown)


stable/labelsmanager-controller/values.yaml
```diff
@@ -0,0 +1,46 @@
...
+  # controller.metricAddr -- Address to serve prometheus metrics on. "0" is disabled.
+  metricAddr: "0"
+
+  # controller.overwrite -- Overwrite any similar label on the pod.
+  overwrite: false
+
+image:
+  repository: thomasnyambati/labelsmanager-controller
+  pullPolicy: IfNotPresent
+  tag: ""
```
max-rocket-internet, line 21: Can you create a tag and put it here?

stable/labelsmanager-controller/values.yaml
```diff
@@ -0,0 +1,46 @@
...
+podSecurityContext: {}
+
+securityContext: {}
+
+extraLabels: {}
+
+service:
+  type: ClusterIP
+  port: 80
+
```
max-rocket-internet, line 39: Not used:

```suggestion
```

stable/labelsmanager-controller/values.yaml
nyambati: I am using the appVersion as default value, we can add this incase we use AWS registry

max-rocket-internet: Ah I see. Nice. Just leave it like this then. I see it's the default behaviour in Helm now.

max-rocket-internet (approved):

PR #94, opened 2020-12-18
[stable/locust]Allow config maps to be mounted in locust helm chart

goelankitt:
## Description

Allow config maps to be mounted in locust helm chart

## Checklist

### DIFF
--- a/stable/locust/values.yaml
+++ b/stable/locust/values.yaml
@@ -102,6 +102,7 @@ ingress:
   #    hosts:
   #      - chart-example.local
 
+configMaps: {}
 nodeSelector: {}
 tolerations: []
 affinity: {}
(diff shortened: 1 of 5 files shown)


stable/locust/values.yaml
```diff
@@ -102,6 +102,7 @@ ingress:
   #    hosts:
   #      - chart-example.local
 
+configMaps: {}
```
max-rocket-internet, line 105: Can we rename this just so it's distinct from the other configmap values? Let's also add a doc item so that readme is updated

```suggestion
#  extraConfigMaps -- Any extra configmaps to mount for the master and worker. Can be used for extra python packages
extraConfigMaps: {}
```

max-rocket-internet (approved):

Issue #108, opened 2021-01-29
After adding "-headless", workers cannot connect to locust master.


lsl-2020:
Hi,

I have deployed this helm chart in AKS cluster. And it works very well. Here is my deployment command in Azure cloud shell:
```
helm install my-loadtest-release ./ \
  --set worker.hpa.enabled=true \
  --set worker.replicas=3 \
  --set service.type=LoadBalancer
```
(Current directory is /helm-charts/stable/locust, and it worked very well with the webpage-mode.)

However, after I simply added "--headless" in master-deployment.yaml, the workers cannot connect to locust master. Here is where I changed:
```
{{ else if .Values.master.args_include_default }}
        args:
          - --master
          - --headless
          - --locustfile=/mnt/locust/{{ .Values.loadtest.locust_locustfile }}
          - --host={{ .Values.loadtest.locust_host }}
          - --loglevel={{ .Values.master.logLevel }}
```

I just added one line "- --headless", and all connection from worker to master failed:
![image](https://user-images.githubusercontent.com/68244035/106224695-c050e500-621e-11eb-9a0d-ffa19c8d9546.png)

What I have tried:
1. I have tried combination of "--run-time", "--spawn-rate", "--run-time", "--expected-workers", all end up with same results.
2. I have tried changing service name from {{ template "locust.fullname" . }} to {{ template "locust.fullname" . }}-master, same with master's name. Still the same.
3. I have tried to manully bind master-port using "--master-bind-port" in master-deployment.yaml, and "--master-host", "--master-port" in worker-deployment.yaml" (In all kinds of combinations). Still the same.
4. I have tried to use Environment variables to do all mentioned changes, and get the same results.

I think it is neccessary to be able to use headless mode for automated load test. So much appreciated if anyone gets curious and help with this issue. Thank you!

My working environments:
- Resource: An AKS (Azure Kubernetes Service) Cluster
- Command portal: Azure Cloud Shell

max-rocket-internet:
Hi @lsl-2020

It's because there is a `readinessProbe` for the master here: https://github.com/deliveryhero/helm-charts/blob/master/stable/locust/templates/master-deployment.yaml#L103-L110

If you use headless mode, the HTTP server isn't started, the `readinessProbe` will fail, the pod will not receive requests via the k8s service.

> I think it is neccessary to be able to use headless mode for automated load test.

I agree.

[...]

PR #109, opened 2021-02-02
[stable/locust] add options for headless

max-rocket-internet:
- Also removing liveliness probe as I don't think it's ever useful to have the master restarted.
- Add separate env vars for master/worker
- Add doc for master/worker `args`

Resolves https://github.com/deliveryhero/helm-charts/issues/108

## Checklist

### DIFF
--- a/stable/locust/Chart.yaml
+++ b/stable/locust/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: locust
-version: "0.9.8"
+version: "0.9.9"
 appVersion: 1.4.1
 home: https://github.com/locustio/locust
 icon: https://locust.io/static/img/logo.png
--- a/stable/locust/README.md
+++ b/stable/locust/README.md
@@ -74,7 +74,7 @@ helm install my-release deliveryhero/locust -f values.yaml
 | ingress.enabled | bool | `false` |  |
 | ingress.hosts[0].host | string | `"chart-example.local"` |  |
 | ingress.tls | list | `[]` |  |
-| loadtest.environment | object | `{}` | environment variables used in the load test |
+| loadtest.environment | object | `{}` | environment variables used in the load test for both master and workers |
 | loadtest.headless | bool | `false` | whether to run locust with headless settings |
 | loadtest.locust_host | string | `"https://www.google.com"` | the host you will load test |
 | loadtest.locust_lib_configmap | string | `""` | name of a configmap containing your lib |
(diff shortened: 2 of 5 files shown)


lsl-2020:
Hi Max, thank you for your work to add headless mode support in the chart.

I have tested this PR, both in non-headless and headless mode. Here is what I did:

# Non-headless mode
```
helm install my-release ./ \
  --set worker.hpa.enabled=true \
  --set worker.replicas=3 \
  --set service.type=LoadBalancer
```
## Web-page statistics:
non-headless_test
## Master node logs:
non-headless_test1

# Headless mode
```
helm install my-release ./ \
  --set worker.hpa.enabled=true \
  --set worker.replicas=3 \
  --set loadtest.headless=true
```
## Master node logs:
headless_test0

# Suggestion for this PR
It feels great to be able to run load test in headless mode. Based on what we have here, we need to support more test configurations binding to `--headless`. The chart now only allow users to do it via environment variables (`loadtest.environment`).

So I think it is really necessary for us to split `loadtest.environment` into `master.environment` and `worker.environment`. In this way, we can support something like this:
```
helm install my-release ./ \
  --set worker.hpa.enabled=true \
  --set worker.replicas=3 \
  --set loadtest.headless=true \
  --set master.environment.LOCUST_SPAWN_RATE=10 \
  --set master.environment.LOCUST_USERS=100 \
  --set master.environment.LOCUST_RUN_TIME=1m \
  --set master.environment.LOCUST_EXPECT_WORKERS=3
```
Acctually `loadtest.headless` can also be replaced by `master.LOCUST_HEADLESS` and `worker.LOCUST_HEADLESS`, but this is purely optional and the current way seems just fine.


Once again great work Max! And I hope this helps.

max-rocket-internet:
> So I think it is really necessary for us to split loadtest.environment into master.environment and worker.environment

OK done

I think you can also achieve what you want by using the `args` instead of env vars, but this wasn't documented in the README, so I've added it.

Let me know if it works 🙂

nreymundo (approved):

lsl-2020:
@max-rocket-internet Hi Max, I have tested it with env vars. Everything looks great.

However I don't know how to use args to install the chart, can you please provide with an example like this?
```
helm install my-release ./ \
  --set worker.hpa.enabled=true \
  --set worker.replicas=3 \
  --set loadtest.headless=true \
  --set master.environment.LOCUST_SPAWN_RATE=10 \
  --set master.environment.LOCUST_USERS=100 \
  --set master.environment.LOCUST_RUN_TIME=1m \
  --set master.environment.LOCUST_EXPECT_WORKERS=3
```

Thank you!

max-rocket-internet:
Hi @lsl-2020 

> I have tested it with env vars. Everything looks great.

Cool, let's merge then.

> I don't know how to use args to install the chart, can you please provide with an example like this?

By args, I meant locust command args. So if you use a chart values file like this:

```yaml
master:
  args:
    - --headless
    - --users=1000
    - --spawn-rate=0.3
```

Then the locust master will be started like this:

```
locust --master - --locustfile=/mnt/locust/xxx --host=xxx --loglevel=xxx --headless --users=1000 --spawn-rate=0.3
```

This is templated here: https://github.com/deliveryhero/helm-charts/blob/master/stable/locust/templates/master-deployment.yaml#L45-L53

All the best!

lsl-2020:
Hi Max,



> By args, I meant locust command args. So if you use a chart values file like this:
> 
> ```yaml
> master:
>   args:
>     - --headless
>     - --users=1000
>     - --spawn-rate=0.3
> ```

I was asking how to specify args when using "--set" in command line. But I guess using values.yaml as you mentioned above is a better way. Btw I have already tested it :-)

Thanks for the great work!

PR #111, opened 2021-02-18
[stable/aws-service-events-exporter] add aws-service-events-exporter

nyambati:
## Checklist


README.md
```diff
@@ -49,6 +49,8 @@ helm install my-release deliveryhero/<chart>
 - [service-account](stable/service-account)
 - [toxiproxy](stable/toxiproxy)
 - [weblate](stable/weblate)
+- [labelsmanager-controller](stable/labelsmanager-controller)
+- [aws-events-exporter](stable/aws-events-exporter)
```
max-rocket-internet, line 53: Can you put this in alphabetical order?

You can run:

```
for chart in $(ls -1 stable/); do echo "- [$chart](stable/$chart)"; done
```

stable/aws-events-exporter/Chart.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v2
+name: aws-events-exporter
+description: "This helm chart exports aws events to prometheus via aws SQS service"
```
max-rocket-internet, line 3: If I understand correctly, it's specifically for RDS events? Like there's no general AWS events service?

If so, I think you either need to rename the chart to `aws-rds-events-exporter` or at least make it very clear in the description and/ore readme that it's for RDS events 🙂

stable/aws-events-exporter/values.yaml
```diff
@@ -0,0 +1,42 @@
...
+  pullPolicy: IfNotPresent
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: 1.0.0
+
+imagePullSecrets: []
+nameOverride: ""
+fullnameOverride: ""
+# serviceAccountName -- service account to be used by the containers
+serviceAccountName: ""
+podAnnotations: {}
```
max-rocket-internet, line 18: If you install this chart as is, prom will never find the `/metrics` endpoint as the default annotations are not present either on the service or on the pod.

I think it's better to include the required pod annotations by default and make the service optional, like here: https://github.com/deliveryhero/helm-charts/blob/master/stable/prometheus-soti-mobicontrol-exporter/values.yaml#L55-L66

stable/aws-events-exporter/values.yaml
```diff
@@ -0,0 +1,42 @@
...
+tolerations: []
+affinity: {}
+
+exporter:
+  # exporter.port -- Exporter port
+  port: 9090
+  # exporter.listenAddress -- Exporter listening address
+  listenAddress: 0.0.0.0
+  # exporter.queueUrl -- SQS queue url
+  queueUrl: ""
```
max-rocket-internet, line 38: Is this optional? Because from the chart description it sounds like it's required? But setting `""` makes it optional?

stable/aws-events-exporter/Chart.yaml
nyambati: Phil suggested we make it generic since we fetch events from sqs queue. Therefore we can be able to use it with other services as well if we need to.

max-rocket-internet: I didn't know there was other events like that but OK, sounds like a good approach!

stable/aws-events-exporter/templates/deployment.yaml
```diff
@@ -0,0 +1,70 @@
...
+      containers:
+        - name: {{ .Chart.Name }}
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          args:
+            - --port={{ .Values.exporter.port }}
+            - --listen-address={{ .Values.exporter.listenAddress }}
+            - --queue-url={{ required "Missing required SQS queue URL" .Values.exporter.queueUrl }}
```
max-rocket-internet, line 39: That's cool 😃

max-rocket-internet (approved):
Well done!

The CI is failing because of the missing `exporter.queueUrl` value so just create a `ct-values.yaml` and set replicas to 0 or something similar.

PR #115, opened 2021-02-24
[stable/prometheus-cloudflare-exporter] Add chart

javad-hajiani:
## Checklist

### DIFF
--- a/stable/prometheus-cloudflare-exporter/templates/service.yaml
+++ b/stable/prometheus-cloudflare-exporter/templates/service.yaml
@@ -7,7 +7,7 @@ metadata:
 spec:
   type: {{ .Values.service.type }}
   ports:
-    - port: 80
+    - port: 9199
       targetPort: 9199
       protocol: TCP
   selector:
--- a/stable/prometheus-cloudflare-exporter/templates/service.yaml
+++ b/stable/prometheus-cloudflare-exporter/templates/service.yaml
@@ -1,9 +1,12 @@
+{{- if .Values.service.create }}
 apiVersion: v1
 kind: Service
 metadata:
   name: {{ include "prometheus-cloudflare-exporter.fullname" . }}
   labels:
-{{ include "prometheus-cloudflare-exporter.labels" . | indent 4 }}
+    {{ include "prometheus-cloudflare-exporter.labels" . | indent 4 }}
+  annotations:
+    {{- toYaml .Values.service.annotations | nindent 8 }}
 spec:
   type: {{ .Values.service.type }}
   ports:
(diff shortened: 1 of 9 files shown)


stable/prometheus-cloudflare-exporter/values.yaml
```diff
@@ -0,0 +1,57 @@
...
+image:
+  repository: osig/prometheus-cloudflare-exporter
+  tag: 1.2.0
+  pullPolicy: IfNotPresent
+
+imagePullSecrets: []
+nameOverride: ""
+fullnameOverride: ""
+
+serviceAccount:
```
nyambati, line 16: You don't have service account config, rename this to just serviceAccountName

max-rocket-internet (changes requested):
You don't have any annotations included for prom exporters, so prom will never find the `/metrics` endpoint.

I think it's better to include the required pod annotations by default and make the service optional, like here: https://github.com/deliveryhero/helm-charts/blob/master/stable/prometheus-soti-mobicontrol-exporter/values.yaml#L55-L66

nyambati (changes requested):
Include pod and service annotations for prometheus 
https://github.com/deliveryhero/helm-charts/blob/3bce1203e9bc24ff4113e8d8b8b2450a80522179/stable/aws-service-events-exporter/values.yaml#L19

max-rocket-internet:
You also need to include the options from here: https://github.com/wehkamp/docker-prometheus-cloudflare-exporter#try-it

For example, I would put `AUTH_KEY` in a secret, like here: https://github.com/deliveryhero/helm-charts/blob/master/stable/prometheus-soti-mobicontrol-exporter/templates/secret.yaml

javad-hajiani:
> You also need to include the options from here: https://github.com/wehkamp/docker-prometheus-cloudflare-exporter#try-it
> 
> For example, I would put `AUTH_KEY` in a secret, like here: https://github.com/deliveryhero/helm-charts/blob/master/stable/prometheus-soti-mobicontrol-exporter/templates/secret.yaml

I've wanted to just pass them as environment variable to deployment which could be more flexible

stable/prometheus-cloudflare-exporter/templates/service.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "prometheus-cloudflare-exporter.fullname" . }}
+  labels:
+{{ include "prometheus-cloudflare-exporter.labels" . | indent 4 }}
+spec:
+  type: {{ .Values.service.type }}
+  ports:
+    - port: 80
```
max-rocket-internet, line 10: I think it's conventional for exporters to use the same port

```suggestion
    - port: 9199
```

max-rocket-internet (approved):
Nice!

javad-hajiani:
> Nice!

Thanks for giving awesome suggestions

nyambati (approved):

PR #119, opened 2021-03-01
[stable/cluster-overprovisioner] Add service account to cluster-overprovisioner chart

joshuastern:
## Description

Add ability to use a non-default service account for the cluster-overprovisioner

## Checklist

### DIFF
--- a/stable/cluster-overprovisioner/Chart.yaml
+++ b/stable/cluster-overprovisioner/Chart.yaml
@@ -8,7 +8,7 @@ description: |
   This approach is the [current recommended method to achieve overprovisioning](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-can-i-configure-overprovisioning-with-cluster-autoscaler).
 name: cluster-overprovisioner
 home: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler
-version: 0.5.0
+version: 0.6.0
 maintainers:
 - name: max-rocket-internet
   email: max.williams@deliveryhero.com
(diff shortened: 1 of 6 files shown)


max-rocket-internet:
Hi @joshuastern, thanks for the PR. Can you fix the docs? Details are here: https://github.com/deliveryhero/helm-charts#contributing

Also, I'm curious, why does `cluster-overprovisioner` need a serviceAccount? It just runs and does nothing, doesn't need to connect to the k8s API etc?

joshuastern:
> Hi @joshuastern, thanks for the PR. Can you fix the docs? Details are here: https://github.com/deliveryhero/helm-charts#contributing
> 
> Also, I'm curious, why does `cluster-overprovisioner` need a serviceAccount? It just runs and does nothing, doesn't need to connect to the k8s API etc?

Exactly, it does not need a service account token. That's why we intend to set the `automountServiceAccountToken` to false when installing the chart. I could set the default value for this to `false`, but it seems better to keep it consistent with the default kubernetes behavior.

max-rocket-internet:
> Exactly, it does not need a service account token. That's why we intend to set the automountServiceAccountToken to false

I still don't get the point though. You just want to have the pods running without the token?

joshuastern:
> > Exactly, it does not need a service account token. That's why we intend to set the automountServiceAccountToken to false
> 
> I still don't get the point though. You just want to have the pods running without the token? Therefore you create a new serviceAccount so that you can set `automountServiceAccountToken` to false?

Yup. In its current state, it runs with a token for the default service account.

max-rocket-internet (approved):
Makes sense. Thanks!

max-rocket-internet:
@joshuastern the CI test for markdown lint has started failing for some reason. I will investigate.

max-rocket-internet:
I opened an issue here: https://github.com/avto-dev/markdown-lint/issues/13

PR #127, opened 2021-03-17
[stable/locust] fix locust static files notfound when use ingress & use locust ver1.4.3

linzhengen:
## Description
- Locust static files(e.g.  `./static/css/application.css?v=1.4.3` ) Not Found when I use ingress
- `locustio/locust:1.4.3` is released 2 months ago [(docker hub)](https://hub.docker.com/r/locustio/locust/tags?page=1&ordering=last_updated)

## Checklist

### DIFF
--- a/stable/locust/Chart.yaml
+++ b/stable/locust/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: locust
-version: "0.9.9"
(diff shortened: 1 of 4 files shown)


max-rocket-internet (approved):
Strange that this would be required 🤔

But OK, thanks @linzhengen!

PR #147, opened 2021-04-20
[stable/locust] add options for tags & excludeTags

linzhengen:
## Description
- added locust with `--tags [TAG [TAG ...]]` options, so only tasks with any matching tags will be executed
- added locust with `--exclude-tags [TAG [TAG ...]]` options, so only tasks with no matching tags will be executed
- fix `docker-entrypoint.sh` doesn't work with list arguments(e.g. `--tags tag1 tag2`)
- locust docs: https://docs.locust.io/en/1.0.3/writing-a-locustfile.html#tagging-tasks

## Checklist

### DIFF
--- a/stable/locust/Chart.yaml
+++ b/stable/locust/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: locust
-version: "0.9.11"
+version: "0.9.12"
 appVersion: 1.4.3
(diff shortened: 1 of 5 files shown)


max-rocket-internet (approved):

max-rocket-internet:
Thanks @linzhengen

PR #172, opened 2021-05-11
[stable/locust] add environment variables from existing secret

jfrancisco0:
## Description

- There is a value `environment_secret` that creates a secret and mounts its keys as environment variables. This was, however, not documented, so I added the details to `values.yaml` and the README.
- If `environment_secret` was not set, an empty secret with name `locust` would be created. I changed this, so the secret is only created if there is data to fill it with.
- While this is great, when using GitOps tools like Flux, the contents of `values.yaml` are uploaded to Git, so this is not a great way to pass secrets to the deployment for these users. To avoid this, I added the `environment_external_secret` value, so an existing secret can be used for environment variables. This way you can create the secret some other way, without keeping its contents in plain text on Git. The format I'm suggesting is:

```yaml
environment_external_secret:
  secret_name: env_var_name
```

Where `secret_name` is the name of the secret already created, not managed by the chart. And `env_var_name` is the name of the environment variable, and also the name of the relevant key in the secret containing the desired secret value.

For example, for the secret created as

```
kubectl create secret generic test-secret \
  --from-literal=token=abcd \
  -n locust
 ```

The value would be
```yaml
environment_external_secret:
  test-secret: token
```

Does this sound good to you?

## Checklist

### DIFF
--- a/stable/locust/Chart.yaml
+++ b/stable/locust/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: locust
-version: "0.9.18"
+version: "0.9.19"
 appVersion: 1.4.4
 home: https://github.com/locustio/locust
 icon: https://locust.io/static/img/logo.png
--- a/stable/locust/README.md
+++ b/stable/locust/README.md
@@ -76,8 +76,8 @@ helm install my-release deliveryhero/locust -f values.yaml
 | ingress.hosts[0].path | string | `"/"` |  |
 | ingress.tls | list | `[]` |  |
 | loadtest.environment | object | `{}` | environment variables used in the load test for both master and workers |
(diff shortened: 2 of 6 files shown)


mshero:
If I'm not mistaken we have our secrets already encrypted in `secrets.yaml` files where necessary and not as plain text in the `values.yaml`. I guess @max-rocket-internet can shed more light on the details

max-rocket-internet (approved):
Great contribution @jfrancisco0, thank you very much 💛

max-rocket-internet:
> If I'm not mistaken we have our secrets already encrypted in secrets.yaml files where necessary and not as plain text

We do, yes. But this is with [helm-secrets](https://github.com/zendesk/helm-secrets), a great tool for sure, but not everyone is using this.

PR #176, opened 2021-05-17
[stable/kubecost-reports-exporter] add serviceaccount

nyambati:
## Checklist

### DIFF
--- a/stable/kubecost-reports-exporter/templates/serviceaccount.yaml
+++ b/stable/kubecost-reports-exporter/templates/serviceaccount.yaml
@@ -1,0 +1,10 @@
+{{- if .Values.serviceAccount.create }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "kubecost-reports-exporter.serviceAccountName" . }}
+  labels:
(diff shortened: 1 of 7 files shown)


bilenkis (approved):

stable/kubecost-reports-exporter/values.yaml
```diff
@@ -14,7 +14,10 @@ image:
 imagePullSecrets: []
 nameOverride: ""
 fullnameOverride: ""
-serviceAccountName: ""
+serviceAccount:
+  enabled: false
```
max-rocket-internet, line 18: But here `enabled`

stable/kubecost-reports-exporter/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,10 @@
+{{- if .Values.serviceAccount.create }}
```
max-rocket-internet, line 1: Here `create` is used

stable/kubecost-reports-exporter/values.yaml
nyambati: I will change this

PR #199, opened 2021-07-06
[stable/locust] Added the ability to enable UI basic auth from the values.yaml

yosri2005:
Added the ability to enable UI basic auth from the values.yaml. The template will include the necessary arg to master as well as adjust the readinessProbe using the given username and password.

Using master.auth.enabled, one is now able to enable basic auth on the Locust UI. This will also require providing master.auth.username and master.auth.password, which will be used to add the required args to enabled web authentication, and to adjust the readinessProbe to use basic auth. 

## Checklist

- [ X] Title of the PR starts with chart name (e.g. `[stable/mychartname]`)
- [ X] I have read the [contribution instructions](https://github.com/deliveryhero/helm-charts#contributing), bumped chart version and regenerated the docs
- [ X] Github actions are passing

### DIFF
--- a/stable/locust/Chart.yaml
+++ b/stable/locust/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: locust
-version: "0.9.19"
+version: "0.1.20"
 appVersion: 1.4.4
 home: https://github.com/locustio/locust
 icon: https://locust.io/static/img/logo.png
--- a/stable/locust/README.md
+++ b/stable/locust/README.md
@@ -1,6 +1,6 @@
 # locust
 
-![Version: 0.9.19](https://img.shields.io/badge/Version-0.9.19-informational?style=flat-square) ![AppVersion: 1.4.4](https://img.shields.io/badge/AppVersion-1.4.4-informational?style=flat-square)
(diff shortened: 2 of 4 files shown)


stable/locust/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: locust
-version: "0.9.19"
+version: "0.1.20"
```
max-rocket-internet, line 3: This would be an older version as 1 is less than 9 🙂

```suggestion
version: "0.19.20"
```

yosri2005: Thanks @max-rocket-internet; I missed that. I committed the suggestion.

max-rocket-internet:
> Github actions are passing

👀

yosri2005:
@max-rocket-internet I do notice a few checks fail, for example:

```helm-conftest running for chart: stable/backstage...
Error: parse error in "backstage/templates/ingress.yaml": template: backstage/templates/ingress.yaml:1: function "urlParse" not defined
Error: Process completed with exit code 1.```

When I ran checks locally, I noticed that there were a few updates to the backstage, cachet, and webplate READMEs, as well as the webplate and cachet requirements.yaml. I didn't include these changes in my commits because they're not related to my change and didn't want to include them without a good reason. Let me know if i should include them.

max-rocket-internet (approved):
Great, thank you @yosri2005 💖

PR #211, opened 2021-07-29
[stable/mlflow] create mlflow chart initially

lena-kuhn:
## Checklist


stable/mlflow/templates/database.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: postgresql.org/v1
```
max-rocket-internet, line 1: This needs to be optional, so let's add a condition around it, something like this:

```suggestion
{{ if .Values.mflow.postgres.PostgresDatabaseEnabled -}}
apiVersion: postgresql.org/v1
```

stable/mlflow/Chart.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v2
+name: mlflow
+description: A Helm chart for MLflow
+type: application
+version: 1.0.0
+home: https://www.mlflow.org/
+appVersion: "latest"
```
max-rocket-internet, line 7: It's bad practice to use `latest` so how about use the latest proper tag, 1.9.1?

https://hub.docker.com/layers/larribas/mlflow/1.9.1/images/sha256-05ccb0b46bfb60ff0f88ed0bb0fa006a6e581e2d537f13ebebfeb98d3a13b90b?context=explore

stable/mlflow/values.yaml
```diff
@@ -0,0 +1,55 @@
+# Default values for mlflow.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+image:
+  repository: larribas/mlflow
+  pullPolicy: IfNotPresent
+  tag: latest
```
max-rocket-internet, line 8:
```suggestion
  tag: 1.9.1
```

stable/mlflow/templates/database.yaml
lena-kuhn: Good point, thanks :)!

stable/mlflow/Chart.yaml
lena-kuhn:
```suggestion
appVersion: 1.9.1
```

stable/mlflow/templates/service.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "mlflow.fullname" . }}
+  labels:
+    {{- include "mlflow.labels" . | nindent 4 }}
+  annotations:
```
nyambati, line 7: Remove this annotation, its no longer relevant

nyambati (changes requested):

stable/mlflow/templates/service.yaml
lena-kuhn: True, good catch! removed

nyambati (approved):

PR #220, opened 2021-08-11
[stable/locust] Move configs to env variables, fix argument inclusion conditions

jfrancisco0:
## Description
This PR makes some changes to the handling of parameters and environment variables in the chart:

- Currently, the [master](https://github.com/deliveryhero/helm-charts/blob/master/stable/locust/templates/master-deployment.yaml) and [worker](https://github.com/deliveryhero/helm-charts/blob/master/stable/locust/templates/worker-deployment.yaml) deployments don't handle arguments as expected.
Setting `Values.master.args` or `Values.workers.args` in the [values.yaml](https://github.com/deliveryhero/helm-charts/blob/master/stable/locust/values.yaml#L63) replaces all the default args, even if `Values.master.args_include_default` is true, which is not the expected outcome according to the comment/README. Also, other values (`Values.loadtest.headless`, `Values.master.auth.enabled`, `Values.loadtest.tags` and `Values.loadtest.excludeTags`) depend on `Values.master.args_include_default` being true, and `Values.master.args` being empty, which is not desirable.

- The `locust_host` is being set redundantly as an arg (`--host`) and as an env (`LOCUST_HOST`)

- Some env var names are [outdated](https://docs.locust.io/en/stable/configuration.html#all-available-configuration-options) (`LOCUST_MASTER_HOST` and `LOCUST_MASTER_PORT`)

- Since command args [override](https://docs.locust.io/en/stable/configuration.html#configuration-file) environment variables, in this PR I suggest passing most arguments to environment variables. This way, the user has more flexibility to override the chart's hardcoded defaults by passing an argument, which will override the chart's default config passed as an environment variable.

## Checklist

### DIFF
--- a/stable/locust/Chart.yaml
+++ b/stable/locust/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: locust
-version: "0.19.24"
+version: "0.19.25"
 appVersion: 1.4.4
 home: https://github.com/locustio/locust
 icon: https://locust.io/static/img/logo.png
--- a/stable/locust/README.md
+++ b/stable/locust/README.md
@@ -1,6 +1,6 @@
 # locust
 
-![Version: 0.19.24](https://img.shields.io/badge/Version-0.19.24-informational?style=flat-square) ![AppVersion: 1.4.4](https://img.shields.io/badge/AppVersion-1.4.4-informational?style=flat-square)
+![Version: 0.19.25](https://img.shields.io/badge/Version-0.19.25-informational?style=flat-square) ![AppVersion: 1.4.4](https://img.shields.io/badge/AppVersion-1.4.4-informational?style=flat-square)
 
 A chart to install Locust, a scalable load testing tool written in Python.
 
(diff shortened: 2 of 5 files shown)


jfrancisco0:
This has some breaking potential because of the name change of the value `worker/master.args_include_default` -> `worker/master.envs_include_default`. If the user turned this off and manually configured replacement arguments, these will override the new default environment variables. If some other setup is used (for example, the user disabled `args_include_default` and configured locust via environment variables or other lower priority means), this change may be breaking.

I didn't increment the chart's major version because I think this may be quite an edge case, but I'm open to comments.

max-rocket-internet:
I like the PR and the thoughtful logic. Yes the current situation is quite confusing I think.

> this change may be breaking

That's OK. Any serious user will pin the chart version and then check chart changes before using a new version.

> I didn't increment the chart's major version

Cool.

max-rocket-internet (approved):

max-rocket-internet:
Thanks @jfrancisco0 💙

PR #231, opened 2021-08-23
[stable/mlflow] fix metadata.name of serviceaccount object

lena-kuhn:
## Checklist

### DIFF
--- a/stable/mlflow/templates/serviceaccount.yaml
+++ b/stable/mlflow/templates/serviceaccount.yaml
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: {{ include "mlflow.serviceAccountName" . }}
+  name: {{ include "mlflow.fullname" . }}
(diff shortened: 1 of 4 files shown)


stable/mlflow/templates/serviceaccount.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: {{ include "mlflow.serviceAccountName" . }}
+  name: {{ include "mlflow.fullname" . }}
```
lena-kuhn, line 5: I guess simply using the mlflow.fullname template should be enough or? Or do we explicitly want to have the ability to have a serviceAccountName different from fullname?

max-rocket-internet: I think `mlflow.fullname` is enough for now for this chart

max-rocket-internet (approved):

PR #237, opened 2021-09-02
mlflow: Adding a fix for using a different value for targetport

tasdikrahman-deliveryhero:
mlflow: Adding a fix for using a different value for targetport

## Checklist

### DIFF
--- a/stable/mlflow/values.yaml
+++ b/stable/mlflow/values.yaml
@@ -17,9 +17,6 @@ service:
   port: 80
   targetPort: 5000
 
-deployment:
-  port: 5000
-
 serviceAccount:
   # serviceAccount.create -- Specifies whether a service account should be created. If true, mlflow.fullname template is used as name.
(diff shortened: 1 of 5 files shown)


gappan (approved):

stable/mlflow/values.yaml
```diff
@@ -14,7 +14,8 @@ extraLabels: {}
 
 service:
   type: ClusterIP
-  port: 5000
+  port: 80
+  targetPort: 5000
```
nyambati, line 18: since the target port and deployment port is the same value i will recommend using only one will reduce possible errors that might be caused if they different values.

nyambati (approved):

stable/mlflow/values.yaml
tasdikrahman-deliveryhero: makes sense, fixed.

PR #238, opened 2021-09-02
[stable/aws-service-events-exporter] Update aws service events exporter image

javad-hajiani:
## Checklist

### DIFF
--- a/stable/aws-service-events-exporter/values.yaml
+++ b/stable/aws-service-events-exporter/values.yaml
@@ -5,7 +5,7 @@
 replicas: 1
 
 image:
-  repository: 940776968316.dkr.ecr.eu-west-1.amazonaws.com/dockerhub-mirror/aws-service-events-exporter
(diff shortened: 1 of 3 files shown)


tasdikrahman-deliveryhero (approved):

stable/aws-service-events-exporter/values.yaml
```diff
@@ -5,10 +5,10 @@
 replicas: 1
 
 image:
-  repository: thomasnyambati/aws-service-events-exporter
```
nyambati, line 8: You should leave this as is since the chart is public. change the image with values instead.

nyambati (changes requested):

stable/aws-service-events-exporter/values.yaml
javad-hajiani: makes sense, will change it.

nyambati (approved):

PR #247, opened 2021-09-13
[stable/listmonk] Initialize Listmonk helmchart

javad-hajiani:
## Checklist


stable/listmonk/templates/deployment.yaml
```diff
@@ -0,0 +1,69 @@
...
+  template:
+    metadata:
+      {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      labels:
+        {{- include "listmonk.selectorLabels" . | nindent 8 }}
+        {{- if .Values.podLabels }}
+					{{ toYaml .Values.podLabels | nindent 8 }}
```
max-rocket-internet, line 23: I think this needs some massaging

stable/listmonk/values.yaml
```diff
@@ -0,0 +1,84 @@
+replicaCount: 2
+
+image:
+  repository: listmonk/listmonk
+  pullPolicy: IfNotPresent
+  # Overrides the image tag whose default is the chart appVersion.
+  # tag: ""
```
max-rocket-internet, line 7: Can we include a tag?

javad-hajiani: I thought by default it would be better to get image version from Chart.yaml, wdyt?

max-rocket-internet: Oh yes, that's better!

max-rocket-internet (approved):

stable/listmonk/templates/deployment.yaml
javad-hajiani: Fixed, Thanks

PR #254, opened 2021-10-05
Fix command to push helm charts

goelankitt:
## Description

Helm push plugin uses cm-push command instead of push command now

## Checklist


max-rocket-internet (approved):
Can you also pin the helm-push plugin version so we don't get surprises again in the future?

PR #265, opened 2021-11-08
Adding chart for prefer-node-descheduler

mmingorance-dh:
## Checklist

### DIFF
--- a/stable/prefer-node-descheduler/values.yaml
+++ b/stable/prefer-node-descheduler/values.yaml
@@ -1,0 +1,67 @@
+# Default values for prefer-node-descheduler.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+# replicaCount -- Number of replicas to run.
+replicaCount: 1
+
+
+image:
+  # image.repository -- The docker repository and image to be used.
+  repository: nginx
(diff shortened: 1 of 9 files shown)


stable/prefer-node-descheduler/values.yaml
```diff
@@ -0,0 +1,67 @@
...
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+# replicaCount -- Number of replicas to run.
+replicaCount: 1
+
+
+image:
+  # image.repository -- The docker repository and image to be used.
+  repository: nginx
```
max-rocket-internet, line 11: I think this is a mistake?

max-rocket-internet (changes requested):

stable/prefer-node-descheduler/values.yaml
mmingorance-dh: Yes and no. I don't have a docker image ready yet and I'm still discussing where the image will be stored (GCR, Dockerhub or any other). I'll update it as soon as I have it.

max-rocket-internet:
Also add the chart to the list in the main README when you are ready 🙂

PR #278, opened 2022-01-24
[stable/locust] Add support for annotations on deployments

hazim1093:
## Description
Added support for specifying annotations on deployments. 
`annotations` is already used for podAnnotations so added  `deploymentAnnotations`. 

They are required when using some tools like [Reloader](https://github.com/stakater/Reloader)


max-rocket-internet (approved):
Nice. Thanks @hazim1093 !

PR #291, opened 2022-03-14
[stable/locust] Don't set worker replicas when hpa is enabled

czchen:
## Description

In current locust chart, the worker `spec.replicas` is set even when `worker.hpa.enabled` is `true`, and it causes problem in GitOps tool [Argo CD](https://argo-cd.readthedocs.io/en/stable/).

By design, Argo CD constantly compares resource in Kubernetes against their definition in git. In this case, Argo CD will try to set `spec.replicas` to `1` to match resource definition in git, however, `HorizontalPodAutoscaling` will try to set `spec.replicas` to properly value according to current worker loading. This discrepancy makes worker unstable.

To fix the issue, locust chart shall not set `spec.replicas` when `worker.hpa.enabled` is `true`.

## Checklist

### DIFF
--- a/stable/locust/Chart.yaml
+++ b/stable/locust/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: locust
-version: "0.26.0"
+version: "0.26.1"
 appVersion: 2.1.0
 home: https://github.com/locustio/locust
 icon: https://locust.io/static/img/logo.png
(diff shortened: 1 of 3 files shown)


max-rocket-internet (approved):
Thanks for your contribution @czchen but you will also need to bump chart version and regenerate the markdown files (read second item in checklist)

czchen:
Hi @max-rocket-internet ,

We have bumped the version, please help to check this is okay now, thanks.

PR #309, opened 2022-05-11
[stable/prometheus-aws-health-exporter] Add podLabels to aws health exporter chart

javad-hajiani:
## Description

- Add podLabels to exporter 

## Checklist

### DIFF
--- a/stable/prometheus-aws-health-exporter/templates/deployment.yaml
+++ b/stable/prometheus-aws-health-exporter/templates/deployment.yaml
@@ -19,6 +19,9 @@ spec:
       labels:
         app.kubernetes.io/name: {{ include "prometheus-aws-health-exporter.name" . }}
         app.kubernetes.io/instance: {{ .Release.Name }}
+        {{- if .Values.podLabels }}
+          {{ toYaml .Values.podLabels | nindent 8 }}
+        {{- end }}
     spec:
     {{- with .Values.imagePullSecrets }}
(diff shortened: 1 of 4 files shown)


stable/prometheus-aws-health-exporter/templates/deployment.yaml
```diff
@@ -19,6 +19,9 @@ spec:
       labels:
         app.kubernetes.io/name: {{ include "prometheus-aws-health-exporter.name" . }}
         app.kubernetes.io/instance: {{ .Release.Name }}
+        {{- if .Values.podLabels }}
+          {{ toYaml .Values.podLabels | nindent 8 }}
```
pciang, line 23: Does the indentation before `{{ toYaml` screw up everything? 🤔 
I was thinking it should be an immediate newline without indentation space.

goelankitt (approved):

stable/prometheus-aws-health-exporter/templates/deployment.yaml
javad-hajiani: nindent is almost identical to indent , but begins a new line. It may seem unnecessary, but look how much more readable it makes the file.

max-rocket-internet: Also CI tests would show an error if the indentation was wrong here 🙂

PR #335, opened 2022-07-01
[stable/mlflow] optional https port

colebaileygit:
## Checklist

### DIFF
--- a/stable/mlflow/templates/service.yaml
+++ b/stable/mlflow/templates/service.yaml
@@ -17,6 +17,12 @@ spec:
       port: {{ .Values.service.port }}
       protocol: TCP
       targetPort: {{ .Values.service.targetPort }}
+{{- if .Values.service.https.enable }}
+    - name: https
+      port: {{ .Values.service.https.port }}
+      protocol: TCP
+      targetPort: {{ .Values.service.targetPort }}
+{{- end }}
 {{- if .Values.service.loadBalancerSourceRanges }}
   loadBalancerSourceRanges:
 {{ toYaml .Values.service.loadBalancerSourceRanges | indent 4 }}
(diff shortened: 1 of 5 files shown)


stable/mlflow/templates/service.yaml
```diff
@@ -17,6 +17,12 @@ spec:
       port: {{ .Values.service.port }}
       protocol: TCP
       targetPort: {{ .Values.service.targetPort }}
+{{- if .Values.service.https.enable }}
+    - name: https
+      port: {{ .Values.service.https.port }}
+      protocol: TCP
+      targetPort: {{ .Values.service.targetPort }}
```
dhpollack, line 24: Is this going to work if it is the same as the http targetPort?

dhpollack (commented):
See comment about targetPort

stable/mlflow/templates/service.yaml
colebaileygit: Based that off of this doc: https://aws.amazon.com/premiumsupport/knowledge-center/terminate-https-traffic-eks-acm/

I hope it works, but might be a good idea to let the targetPorts be different as well

colebaileygit: I'll try to find a way to test locally without merging this to master to short-circuit this cycle time 🤔

dhpollack: Hmm, that link seems pretty convincing.  I was just making sure this was intentional and not a copy-paste oversight.

max-rocket-internet: > I'll try to find a way to test locally without merging this to master

If you are using helmfiles, you can simply change the source like this:

```yaml
- name: chart-release-name
  # chart: deliveryhero/mlflow
  chart: /path/to/your/local/chart/copy
  version: 7.6.21
```

dhpollack (approved):
if you the targetPort thing is ok, then :+1:

stable/mlflow/templates/service.yaml
colebaileygit: Thanks! Testing this way was quite useful. Since the deployment only has 1 possible port, I won't make the https block customizable

PR #349, opened 2022-07-29
[stable/datadog-controller] update source

max-rocket-internet:
We have updated chartmuseum to latest version so I want to test that sync still works. So here I'll update this chart with its new source.

### DIFF
--- a/stable/datadog-controller/Chart.yaml
+++ b/stable/datadog-controller/Chart.yaml
@@ -18,9 +18,9 @@ description: |
     type: "query alert"
     message: Service my-service has a high error rate on env:stg
   ```
-version: "1.0"
+version: "1.1"
 appVersion: "0.1"
-home: https://github.com/deliveryhero/datadog-controller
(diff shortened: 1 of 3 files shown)


sietevecesmal (approved):

stable/datadog-controller/Chart.yaml
```diff
@@ -31,4 +31,4 @@ keywords:
 - datadog
 - controller
 sources:
-- https://github.com/deliveryhero/datadog-controller
```
sietevecesmal, line 34: is changing `deliveryhero` for `max-rocket-internet` expected?

max-rocket-internet: Yes. I wrote it but in the end we never used it so it was just sitting under the DH org, unused. So I moved the repo to my own Github as I don't want some non-used, semi abandoned software left under DH org.

sietevecesmal: 🙌🏼

PR #293, opened 2022-03-22
[stable/kube-downscaler] sync cluster role with upstream

runningman84:
same rules as upstream project:
https://codeberg.org/hjacobs/kube-downscaler/src/branch/main/deploy/rbac.yaml

## Checklist

- [x] Title of the PR starts with chart name (e.g. `[stable/mychartname]`)
- [x] I have read the [contribution instructions](https://github.com/deliveryhero/helm-charts#contributing), bumped chart version and regenerated the docs
- [ ] Github actions are passing

PR #351, opened 2022-08-08
[stable/kube-downscaler]: match image.tag with appVersion

hairmare:
## Description

The appVersion in kube-downscaler was updated in #293 but the `image.tag` value still deploys the `21.2.0` version. This updates `image.tag` to match `appVersion`.

## Checklist

## Additional Notes

It might be preferable to use the `default` hack from `helm create` so we only need to set the `appVersion`. I can prepare such a PR if that should be preferred.

There is also a newer version 22.7.1 on codeberg, in the spirit of atmic changes, this only fixes releasing the current `appVersion` but i'll prepare a PR to bump downscaler as well once this is fixed.

### DIFF
--- a/stable/kube-downscaler/Chart.yaml
+++ b/stable/kube-downscaler/Chart.yaml
@@ -1,6 +1,6 @@
 name: kube-downscaler
 apiVersion: v1
-version: "0.5.1"
+version: "0.5.2"
 appVersion: 22.2.0
 description: Scale down Kubernetes deployments after work hours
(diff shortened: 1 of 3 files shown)


max-rocket-internet (approved):
Cool, thanks for fixing @hairmare 🎉

hairmare:
thanks for the quick turnaround, @max-rocket-internet :+1: 

i'll follow up on my notes asap

PR #363, opened 2022-08-16
[stable/locust] Bump Locust version to 2.11.0

cyberw:
Updates Locust to 2.11.0

I was unable to run the GH actions in my fork, but I'm guessing they will run when I make this PR...


max-rocket-internet (approved):
Nice, thanks @cyberw  🚀

PR #371, opened 2022-08-26
[stable/kube-downscaler] Fix clusterrole for argocd rollouts

vmozghovyi:
## Description

Fix clusterrole for argocd rollouts

### DIFF
--- a/stable/kube-downscaler/templates/clusterrole.yaml
+++ b/stable/kube-downscaler/templates/clusterrole.yaml
@@ -27,6 +27,16 @@ rules:
   - update
   - patch
 - apiGroups:
+    - argoproj.io
+  resources:
+    - rollouts
+  verbs:
+    - get
+    - watch
+    - list
+    - update
+    - patch
+- apiGroups:
   - autoscaling
   resources:
   - horizontalpodautoscalers
--- a/stable/kube-downscaler/Chart.yaml
+++ b/stable/kube-downscaler/Chart.yaml
@@ -1,6 +1,6 @@
 name: kube-downscaler
 apiVersion: v1
-version: "0.5.5"
+version: "0.5.6"
 appVersion: 22.7.1
 description: Scale down Kubernetes deployments after work hours
 keywords:
--- a/stable/kube-downscaler/README.md
+++ b/stable/kube-downscaler/README.md
@@ -1,6 +1,6 @@
 # kube-downscaler
 
-![Version: 0.5.5](https://img.shields.io/badge/Version-0.5.5-informational?style=flat-square) ![AppVersion: 22.7.1](https://img.shields.io/badge/AppVersion-22.7.1-informational?style=flat-square)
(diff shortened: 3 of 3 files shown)


stable/kube-downscaler/templates/clusterrole.yaml
```diff
@@ -26,6 +26,16 @@ rules:
...
+- apiGroups:
+    - argoproj.io
+  resources:
+    - rollouts
+  verbs:
+    - get
+    - watch
+    - list
+    - update
+    - patch
```
max-rocket-internet, line 38: Hi @vmozghovyi, thanks for the contribution but I think since Argo is a third party company then it makes sense to move these extra permissions outside of the chart. Could you update the PR to do something like:

1. Add an extra value, something like `rbac.extraRules: []`
2. Add something like this in this file:

```yaml
    {{- if .Values.rbac.extraRules }}
      {{- toYaml .Values.rbac.extraRules | nindent 2 }}
    {{- end }}
```

And then for you, you can set `rbac.extraRules` in your values file to:

```yaml
rbac:
  extraRules:
    - apiGroups:
        - argoproj.io
      resources:
        - rollouts
      verbs:
        - get
        - watch
        - list
        - update
        - patch
```

Make sense?

vmozghovyi: Hi, @max-rocket-internet thanks for your mention.

Yes, that might make sense. But the fact is that you are using the original container. The functionality of which added the ability to work with rollouts. And your templates don't match the original manifests and possible launch arguments.

Look at:
https://codeberg.org/hjacobs/kube-downscaler

> --include-resources
> Downscale resources of this kind as comma separated list. [deployments, statefulsets, stacks, horizontalpodautoscalers, rollouts, scaledobjects] (default: deployments)

https://codeberg.org/hjacobs/kube-downscaler/src/commit/f2f8282e90f1c673e4012f1dfa16e4ad1e1185ff/deploy/rbac.yaml#L32

And what about? https://github.com/deliveryhero/helm-charts/blob/4a046d027a1cda5a18f624e6e5627bcd6fe74f1c/stable/kube-downscaler/templates/clusterrole.yaml#L50

vmozghovyi: But anyway I have updated PR: https://github.com/deliveryhero/helm-charts/pull/382

max-rocket-internet: > But the fact is that you are using the original container. The functionality of which added the ability to work with rollouts.

Cool, but the solution here is better 😉

max-rocket-internet: > https://codeberg.org/hjacobs/kube-downscaler/src/commit/f2f8282e90f1c673e4012f1dfa16e4ad1e1185ff/deploy/rbac.yaml#L32

It's also not a chart, it's just static YAML, so is more primitive.

PR #380, opened 2022-09-13
[stable/k8s-resources] Ingress API version networking.k8s.io/v1 support, add Namespaces

### LINKED ISSUE #377: [stable/k8s-resources] Ingress `networking.k8s.io/v1` support
Current version of https://github.com/deliveryhero/helm-charts/blob/master/stable/k8s-resources/templates/ingress.yaml template doesn't really support up-to-date Ingress API version `networking.k8s.io/v1`. API's spec changes are significant enough to not fit into current `.hosts[].paths[]` schema. Also support of `ingressClassName` is missing too.

Adding support new API version requires either:
- breaking values compatibility with previous chart version
- maintaining complex `.Capabilities.kubeVersion` checks, like done in default helm template https://github.com/helm/helm/pull/9621 (though it's incomplete, no support of `...backend.service.port.name` field)

If you're okay with breaking compatibility, I can open a PR. I'm thinking about dropping support of `networking.k8s.io/v1beta1` entirely. But if you want to maintain compatibility, please consider this as a feature request.

z0rc:
Fixes #377.

Add `defaultBackend` and `ingressClassName` values.

Switch from `hosts` values to `rules`. Rule can be defined without host and
also it better matches Ingress spec.

Also fix `secretName` indentation, it should be nested in `.spec.tls`.

Add ability to create Namespaces.

## Checklist

### DIFF
--- a/stable/k8s-resources/Chart.yaml
+++ b/stable/k8s-resources/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-version: 0.5.5
+version: 0.6.0
 appVersion: 0.0.1
 name: k8s-resources
 description: |
--- a/stable/k8s-resources/Chart.yaml
+++ b/stable/k8s-resources/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-version: 0.6.0
+version: 0.6.1
 appVersion: 0.0.1
 name: k8s-resources
 description: |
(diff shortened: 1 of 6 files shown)


z0rc:
Side note: there are deprecation warnings for other resources, see https://github.com/deliveryhero/helm-charts/actions/runs/3044856406/jobs/4905706751#step:7:33, so testing and deployment will fail with kubenetest 1.25+.

max-rocket-internet (approved):
Great, thanks @z0rc!

Issue #383, opened 2022-09-17
Add configurable resources to node-local-dns


kthompson23:
Thank you for publishing the node-local-dns chart to Artifact Hub. The chart has resource requests defaults but no option to configure resource limits.

Would you accept a PR to make resource requests and limits configurable? I would leave the default request values already specified.

max-rocket-internet:
> Would you accept a PR to make resource requests and limits configurable? 

@kthompson23 of course we will, it would be very welcome 🙂

kthompson23:
@max-rocket-internet Sorry for the delay, I submitted a PR.

max-rocket-internet:
Resolved in https://github.com/deliveryhero/helm-charts/pull/386

PR #386, opened 2022-09-21
[stable/node-local-dns] Add support for configurable resource requests and limts

kthompson23:
## Description

This PR updates the node-local-dns chart to make the resource requests and limits configurable. The resource requests default to `25m` and `5Mi`, which is unchanged from the previous hard-coded values.

Limits is set to an empty object so by default the limits will not be set. Users can choose to set just the cpu, memory, or both.

## Checklist

- [X] I have read the [contribution instructions](https://github.com/deliveryhero/helm-charts#contributing), bumped chart version and regenerated the docs

### DIFF
--- a/stable/node-local-dns/Chart.yaml
+++ b/stable/node-local-dns/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: node-local-dns
-version: 0.1.1
+version: 0.2.0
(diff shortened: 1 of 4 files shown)


kthompson23:
closes #383

max-rocket-internet (approved):
OK nice @kthompson23 🎉

PR #390, opened 2022-10-04
Facing issue while pod creation.

pushpamahadeva:
## Description

Secret having api_token as a key, but in deployment.yaml we are looking api_key.
Wrong call of token, due to this is pod not coming up.

## Checklist


max-rocket-internet:
Hi @pushpamahadeva 
You haven't read the contribution instructions, this PR is not mergeable.

Also this change would break it for existing users for no good reason, you can just set `existingSecret.tokenKey: api_token` to solve your problem 🙂

Issue #385, opened 2022-09-20
[stable/cachet:1.3.1] Duplicated definition of DB_DRIVER env variable


cezaryulita:
Having configmap.yaml defined as
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ include "cachet.fullname" . }}
  labels:
{{ include "cachet.labels" . | indent 4 }}
data:
  # database
  DB_HOST: {{ .Values.database.host  }}
  DB_DATABASE: {{ .Values.database.name  }}
  DB_USERNAME: {{ .Values.database.username }}
  DB_DRIVER: {{ .Values.database.driver }}
  DB_PORT: {{ .Values.database.port | quote }}
  APP_ENV: {{ default "production" .Values.env.APP_ENV | quote }}
  APP_URL: {{ .Values.ingress.host }}
  {{- range $key, $value := .Values.env.public }}
  {{ $key }}: {{ $value | quote }}
  {{- end }}
```
effectively results in release being undeployable by rendering output as
```
data:
  # database
  DB_HOST: cachet-db.example.com
  DB_DATABASE: cachet
  DB_USERNAME: cachet
  DB_DRIVER: pssql
  DB_PORT: "5432"
  APP_ENV: "production"
  APP_URL: chart-example.local
  APP_DEBUG: "false"
  APP_LOG: "errorlog"
  CACHET_BEACON: "false"
  CACHET_EMOJI: "false"
  CACHE_DRIVER: "database"
  DB_DRIVER: "pgsql"
```
that causes:
Helm install failed: error while running post render on files: map[string]interface {}(nil): yaml: unmarshal errors: line 28: mapping key "DB_DRIVER" already defined at line 19

Becram:
@cezaryulita  https://github.com/deliveryhero/helm-charts/pull/393

PR #393, opened 2022-10-15
[stable/chachet] remove DB_DRIVER duplication in configmap

Becram:
## Description

Since the variable is already defined at `database.driver` it causes duplication error in the configmap

## Checklist


Becram:
as raised by https://github.com/deliveryhero/helm-charts/issues/385

max-rocket-internet (approved):
Thanks @Becram but you also need to generate the chart meta data from the templates 🙂

PR #394, opened 2022-10-17
[stable/cloudhealth-collector] Use correct `secretKeyRef.key` for API Token secret

jacobboykin:
## Description

* If a user installs this chart and sets the [`api_token` value](https://github.com/deliveryhero/helm-charts/blob/master/stable/cloudhealth-collector/values.yaml#L1), the cloudhealth-colllector pod will fail to start due to the deployment config using the wrong secret key: 
* https://github.com/deliveryhero/helm-charts/blob/master/stable/cloudhealth-collector/templates/deployment.yaml#L50
* https://github.com/deliveryhero/helm-charts/blob/master/stable/cloudhealth-collector/templates/secret.yaml#L10

## Checklist

### DIFF
--- a/stable/cloudhealth-collector/templates/secret.yaml
+++ b/stable/cloudhealth-collector/templates/secret.yaml
@@ -7,5 +7,5 @@ metadata:
     {{- include "cloudhealth-collector.labels" . | nindent 4 }}
 type: Opaque
 data:
-  api_token: {{ .Values.api_token | b64enc | quote }}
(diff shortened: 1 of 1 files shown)


max-rocket-internet:
Hi @jacobboykin 👋
Since the real environment variable that is used by Cloudhealth is called `CHT_API_TOKEN`, how about we rename all references to `api_token`? I think this would be more logical and consistent, right?

Also, you need to bump the chart version and update the chart readme from the templates 🙂

PR #423, opened 2022-12-19
[stable/locust] fix labels for PodDisruptionBudget and enable for worker

max-rocket-internet:
The labels on the `PodDisruptionBudget` were incorrect and did not match the actual pods. This causes the `cluster-autoscaler` to move the pods on node scale down with events like `deleting pod for node scale down`. And this interrupts running load tests.


mshero (approved):
:rocket:

brunogf (approved):

PR #437, opened 2023-01-25
[stable/weblate] update ingress API

mkamel93:
## Checklist

### DIFF
--- a/stable/weblate/templates/ingress.yaml
+++ b/stable/weblate/templates/ingress.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.ingress.enabled }}
+{{- if .Values.ingress.enabled -}}
 {{- $fullName := include "weblate.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
 {{- if semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion -}}
--- a/stable/weblate/templates/ingress.yaml
+++ b/stable/weblate/templates/ingress.yaml
@@ -1,7 +1,13 @@
 {{- if .Values.ingress.enabled }}
 {{- $fullName := include "weblate.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
+{{- if semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion -}}
 apiVersion: networking.k8s.io/v1
+{{- else if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+apiVersion: networking.k8s.io/v1beta1
+{{- else -}}
+apiVersion: extensions/v1beta1
+{{- end }}
 kind: Ingress
 metadata:
   name: {{ include "weblate.fullname" . }}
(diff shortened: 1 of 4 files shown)


stable/weblate/values.yaml
```diff
@@ -79,14 +79,17 @@ service:
...
   annotations:
     {}
     # kubernetes.io/ingress.class: nginx
     # kubernetes.io/tls-acme: "true"
   hosts:
     - host: chart-example.local
-      paths: []
+      paths:
+        - path: /
+          pathType: ImplementationSpecific
```
mKeRix, line 92:
```suggestion
          pathType: Prefix
```

Prefix feels like a more stable default to me, it's more strict in what it should represent so the implementation across controllers should be more consistent.

stable/weblate/values.yaml
```diff
@@ -79,14 +79,17 @@ service:
   port: 80
 
 ingress:
-  enabled: false
+  enabled: true
```
mKeRix, line 82: Why is the default switched from `false` to `true`?

stable/weblate/values.yaml
```diff
@@ -79,14 +79,17 @@ service:
...
+  className: ""
```
mKeRix, line 83: This value was added, but it doesn't seem to be used anywhere?

stable/weblate/templates/ingress.yaml
```diff
@@ -1,7 +1,7 @@
 {{- if .Values.ingress.enabled }}
 {{- $fullName := include "weblate.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
-apiVersion: networking.k8s.io/v1beta1
+apiVersion: networking.k8s.io/v1
```
mKeRix, line 4: Since you support both old and new API versions in the spec below I would also make this line conditional to support both

mKeRix (changes requested):

stable/weblate/values.yaml
mkamel93: it was changed to test new ingress changes in ci checks as false will neglect creating ing resources

stable/weblate/templates/ingress.yaml
mkamel93: old version support is added

stable/weblate/values.yaml
mkamel93: added using this var in ingress

stable/weblate/values.yaml
mkamel93: reverted after testing

stable/weblate/values.yaml
mkamel93: done

mkamel93: ImeplementationSpecific gives more freedom as its matching is up to the IngressClass. Implementations can treat this as a separate pathType or treat it identically to Prefix or Exact path types.

stable/weblate/values.yaml
mkamel93: reverted to false after testing completes successfully

stable/weblate/values.yaml
mKeRix: I know, which is why I thought using `Prefix` would be more consistent. But it's not a hill that I'll die on, both is ok. It's just a default anyway.

mKeRix (approved):

albrechtsimon (approved):

PR #451, opened 2023-03-07
[stable/node-local-dns] update docker image tag and add memory limit

max-rocket-internet:
Also updating main README with missing chart


stable/node-local-dns/values.yaml
```diff
@@ -1,7 +1,7 @@
 image:
   repository: k8s.gcr.io/dns/k8s-dns-node-cache
   # Overrides the image tag whose default is the chart appVersion.
-  tag: "1.21.1"
+  tag: "1.22.20"
```
mshero, line 4: As we're updating the image tag, shouldn't we also consider updating the chart's [appVersion above](https://github.com/deliveryhero/helm-charts/pull/451/files#diff-9635cbbf3a9f58777d50a18a72b9dd5177a7a2a20cce4b9c3315fc4bc9cb7c72R4)?

max-rocket-internet: Yes, you're right @mshero!

mshero (approved):
:heart_eyes:

mtahaahmed (approved):

PR #453, opened 2023-03-13
[stable/aws-ebs-csi-driver] update chart from upstream

max-rocket-internet:
## Checklist


mtahaahmed (approved):

mKeRix (approved):
Is there any reason to keep maintaining this copy of the Helm chart vs. just redirecting to the [upstream repo](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/install.md#helm)? For users it should not be an issue to switch, from what I gathered the chart in this repo seems to be a carbon copy?

max-rocket-internet:
> Is there any reason to keep maintaining this copy of the Helm chart vs. just redirecting to the [upstream repo](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/install.md#helm)?

I think not. Maybe we make this version the last update. I think when the version here was created there was no repo for the upstream one.

PR #455, opened 2023-03-15
[stable/kube-bench] Update kube-bench image, code cleanup, aks support

Mahagon:
### Changes

* bump kube-bench chart version to 0.1.7
* add aks support
* update cis benchmark for gke to 1.2.0
* update cis benchmark for eks to 1.1.0
* update kube-bench image to v0.6.12
* make volumes and volumeMounts configurable
* make jobname configurable to enable multiple deployments of the cronjob with different settings
* fix indent for container resources
* fix cronjob.command has no value if not overwritten by the deployment
* set backoffLimit to 0 to avoid reruns when kube-bench fails
* bump api chart version to v2

### Helmfile example for aks

I would like to use this chart to deploy kube-bench into our AKS Cluster.
This required different volume mounts and a different benchmark.

```yaml
---
repositories:
  - name: deliveryhero
    url: https://charts.deliveryhero.io/

releases:
  - name: kube-bench-aks
    namespace: kube-bench
    chart: /home/daniel/src/deliveryhero-helm-charts/stable/kube-bench
    values:
      - jobName: kube-bench-aks
      - cronjob:
          command:
            - kube-bench
            - --benchmark
            - aks-1.0
            - run
            - --targets
            - controlplane,managedservices,master,node,policies
      - provider: aks
      - volumeMounts:
          - name: var-lib-kubelet
            mountPath: /var/lib/kubelet
            readOnly: true
          - name: etc-systemd
            mountPath: /etc/systemd
            readOnly: true
          - name: etc-default
            mountPath: /etc/default
            readOnly: true
          - name: etc-kubernetes
            mountPath: /etc/kubernetes
            readOnly: true
          - name: usr-local-bin
            mountPath: /usr/local/mount-from-host/bin
            readOnly: true
      - volumes:
          - name: var-lib-kubelet
            hostPath:
              path: "/var/lib/kubelet"
          - name: etc-systemd
            hostPath:
              path: "/etc/systemd"
          - name: etc-default
            hostPath:
              path: "/etc/default"
          - name: etc-kubernetes
            hostPath:
              path: "/etc/kubernetes"
          - name: usr-local-bin
            hostPath:
              path: "/usr/local/bin"

```

## Checklist

### DIFF
--- a/stable/kube-bench/templates/cron.yaml
+++ b/stable/kube-bench/templates/cron.yaml
@@ -1,12 +1,15 @@
+---
 apiVersion: {{ template "cronjob.apiVersion" . }}
 kind: CronJob
 metadata:
-  name: kube-bench
+  name: {{ template "kube-bench.fullname" . }}
+  labels: {{ include "kube-bench.labels" . | nindent 4 }}
 spec:
   schedule: {{ quote .Values.cronjob.schedule }}
   concurrencyPolicy: "{{ .Values.concurrencyPolicy }}"
   jobTemplate:
     spec:
+      backoffLimit: 0
       template:
         spec:
           {{- if .Values.serviceAccount.create }}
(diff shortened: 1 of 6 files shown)


stable/kube-bench/templates/cron.yaml
```diff
@@ -1,12 +1,14 @@
+---
 apiVersion: {{ template "cronjob.apiVersion" . }}
 kind: CronJob
 metadata:
-  name: kube-bench
+  name: {{ .Values.jobName }}
```
max-rocket-internet, line 5: It's more conventional to use a templated value here:

```suggestion
  name: {{ include "test.fullname" . }}
```

You would needs to update `stable/kube-bench/templates/_helpers.tpl` also with the function.

You can copy the output from `helm create example-chart` and see how it should be done. Would that be OK?

max-rocket-internet (changes requested):

stable/kube-bench/templates/cron.yaml
```diff
@@ -15,41 +17,23 @@ spec:
           hostPID: true
           restartPolicy: Never
           containers:
-            - name: kube-bench
+            - name: {{ .Values.jobName }}
```
max-rocket-internet, line 20: Here also the convention is:

```suggestion
            - name: {{ .Chart.Name }}
```

stable/kube-bench/templates/cron.yaml
Mahagon: good point, added
i also added the labels from the example chart
image

stable/kube-bench/templates/cron.yaml
Mahagon: added :)

max-rocket-internet (approved):
Amazing! Thank you very much @Mahagon 💛

PR #459, opened 2023-03-20
[stable/locust] allow to configure nodeSelector, tolerations, affinity for worker explicitly

zifter:
## Description
Allow to configure nodeSelector, tolerations, affinity for worker explicitly.
It will be very useful where master is running using default node pool (so, no need to specify any toleration and affinity), but workers are needed to run on spot instances, which is in another node pool with specific taints. So, in that case, we need to configure only worker deployment

## Checklist

### DIFF
--- a/stable/locust/Chart.yaml
+++ b/stable/locust/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: locust
-version: "0.31.2"
+version: "0.31.3"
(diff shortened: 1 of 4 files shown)


zifter:
@max-rocket-internet, hi!
Can you help with review this pr?

max-rocket-internet (approved):
Nice, thanks @zifter 💛

PR #463, opened 2023-04-11
[stable/node-local-dns] Add Service Monitor, Grafana Dashboard, Image Pull Secret support

treksler:
## Description

- Add Service Monitor, Grafana Dashboard, Image Pull Secret support.
- Renamed the service to remove -prometheus suffix (was breaking Grafana dashboards)
- Renamed upstream service file
- Renamed config variables to follow helm naming conventions. Hence major version number bump

## Checklist


stable/node-local-dns/values.yaml
```diff
@@ -26,6 +27,19 @@ serviceAccount:
 
 podAnnotations: {}
 
+# https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/getting-started.md
+serviceMonitor:
+  # Ensure that servicemonitor is created
+  enabled: false
+
+dashboard:
```
max-rocket-internet, line 35: Can you add a link here to docs about this grafana sidecar?

```suggestion
# <Some details here>
dashboard:
```

max-rocket-internet (approved):
Love it, thanks @treksler 💛

But you need to fix a couple minor things for CI tests.

stable/node-local-dns/values.yaml
treksler: grafana has some documentation in their helm chart README, but it is not immediately obvious how to use it.
- https://github.com/grafana/helm-charts/blob/main/charts/grafana/README.md

the grafana documentation here https://grafana.com/docs/grafana/latest/dashboards is very point and click oriented and makes no mention of this feature that i can find

some other completely unrelated projects have written much better documentation on how to use this feature
 - https://docs.d2iq.com/dkp/2.3/grafana-dashboard
 - https://ranchermanager.docs.rancher.com/how-to-guides/advanced-user-guides/monitoring-alerting-guides/create-persistent-grafana-dashboard

also, for example, kube-prometheus-stack uses this feature to load its built in dashboards, as does kubecost
 - https://github.com/kubecost/cost-analyzer-helm-chart/tree/develop/cost-analyzer/templates
 - https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack/templates/grafana/dashboards-1.14
 
 I would be tempted to link to https://github.com/grafana/helm-charts/blob/main/charts/grafana/README.md but not sure how useful it would be. 
 
 basically googling `grafana_dashboard: "1"` will get the best results :D

max-rocket-internet: OK cool, thanks for the info

PR #474, opened 2023-05-31
[stable/node-local-dns] Allow health endpoint port override

eliasbokreta:
## Description

Hello, here's a small change to allow overriding the `health` endpoint port in some cases where this port is already used by other components.
As an example, Kured is also using `8080` and thus `NodeLocal DNSCache` is failing :
```bash
listen tcp 169.254.20.20:8080: bind: address already in use
```
```bash
$> sudo netstat -nltp | grep 8080
tcp6       0      0 :::8080                 :::*                    LISTEN      3839/kured
```

The modification is currently running fine in one of our k8s clusters.

## Checklist

- [X] I have read the [contribution instructions](https://github.com/deliveryhero/helm-charts#opening-a-pr), bumped chart version and regenerated the docs
- [X] Github actions are passing

### DIFF
--- a/stable/node-local-dns/Chart.yaml
+++ b/stable/node-local-dns/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: node-local-dns
-version: 1.0.0
+version: 1.0.1
 appVersion: 1.22.20
 maintainers:
 - name: gabrieladt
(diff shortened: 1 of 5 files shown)


max-rocket-internet (approved):
Cool, thanks @eliasbokreta 💚

PR #481, opened 2023-06-13
[stable/aws-s3-proxy]: Support k8s 1.19 Ingress

BlueHorn07:
## Description

Ingress of `aws-s3-proxy` only support `networking.k8s.io/v1beta1` and `extensions/v1beta1`, so it cannot be deployed over k8s 1.19 versions.

So, add if-else statements to support `networking.k8s.io/v1` apiVersion. This change is compatible to k8s 1.19 under version, so don't worry :)

## Checklist

### DIFF
--- a/stable/aws-s3-proxy/templates/ingress.yaml
+++ b/stable/aws-s3-proxy/templates/ingress.yaml
@@ -18,6 +18,9 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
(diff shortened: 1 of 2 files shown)


max-rocket-internet (approved):
I would say don't even both with these conditional logic, a new version of the chart will be created so there's no need for backwards compatibilty but all good! Thanks @BlueHorn07 💙

BlueHorn07:
Oh, I didn't bumped chart version! I will bump it to `0.1.4` :pray:

PR #492, opened 2023-07-03
[stable/node-local-dns] feat: Add option to specify protocol used for upstream

zzorica:
This will allow to specify protocol used for upstream communication. CoreDNS is sometimes tricky to adjust on Cloud Providers so it would be easier to specify it here as well. Especially current hardcoded value is tricky as in case if both force_tcp and prefer_udp options are specified the force_tcp takes precedence.

More info [here](https://coredns.io/plugins/forward/)

## Checklist


max-rocket-internet (approved):
Awesome, thanks @zzorica 💕

PR #493, opened 2023-07-05
[stable/node-local-dns] feat: Add option to set securityContext

zzorica:
Node-local-dns doesn't need to run in privileged mode and NET_ADMIN cap is enough. But just in case I am missing something this would set it as an option and leave privileged mode as "default".

To run in non-privileged mode in values this will do it:
```
securityContext:
  capabilities:
    add:
    - NET_ADMIN
```

## Checklist

### DIFF
--- a/stable/k8s-event-logger/Chart.yaml
+++ b/stable/k8s-event-logger/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
-appVersion: "2.0"
-version: "1.1.4"
+appVersion: "2.1"
+version: "1.1.5"
 description: |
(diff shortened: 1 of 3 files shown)


max-rocket-internet (approved):
OK cool, thanks @zzorica

stable/node-local-dns/values.yaml
```diff
@@ -33,6 +33,9 @@ serviceAccount:
 
 podAnnotations: {}
 
+securityContext:
+  privileged: true
```
max-rocket-internet, line 37: It would probably be better to have this default to `false`, no?

max-rocket-internet, line 37: It would probably be better to have this default to `false`, no?

```suggestion
  privileged: false
```

PR #497, opened 2023-07-19
[stable/node-local-dns] Disable prometheus annotations when servicemonitor is enabled

mbugeia-tf1:
## Description

Prometheus annotations for scraping should not be present when using servicemonitor. I disable them when servicemonitor is enabled.

My use-case is:
- We are in the process in switching from annotation based scraping to prometheus operator with servicemonitor.
- We have to maintain annotation based scraping configuration to support services not yet migrated.
- If there are both annotations and servicemonitor, metrics are scrapped 2 times.

## Checklist

### DIFF
--- a/stable/node-local-dns/README.md
+++ b/stable/node-local-dns/README.md
@@ -1,6 +1,6 @@
 # node-local-dns
 
-![Version: 1.1.2](https://img.shields.io/badge/Version-1.1.2-informational?style=flat-square) ![AppVersion: 1.22.20](https://img.shields.io/badge/AppVersion-1.22.20-informational?style=flat-square)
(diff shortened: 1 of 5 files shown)


stable/node-local-dns/README.md
```diff
@@ -12,37 +12,11 @@ Further documentation is [here](https://kubernetes.io/docs/tasks/administer-clus
 
 This helm chart works for both kube-proxy setups (iptables or ipvs).
 
-## How to install this chart
```
max-rocket-internet, line 15: The tests are failing because you're removing all this. Do you need help with that?

mbugeia-tf1: it's fixed, I was confused because I did run the automatic generation of readme but in the chart directory, not at the root of the repo

max-rocket-internet (approved):
Great, thanks @mbugeia-tf1 💙

PR #503, opened 2023-08-08
stable/node-local-dns Extend daemonset arg list

kkonnov:
## Description

Extend daemonset arg list

## Checklist

- [ ] Title of the PR starts with chart name (e.g. `[stable/mychartname]`)
- [ ] I have read the [contribution instructions](https://github.com/deliveryhero/helm-charts#opening-a-pr), bumped chart version and regenerated the docs
- [ ] Github actions are passing

PR #510, opened 2023-08-29
[stable/node-local-dns] Changes for cilium local redirect support

STASiAN:
## Description
-  Return default configuration to be compatible  with iptables setup.
- add dnsServer to -localip only if setupIptables is true
- bind to dnsServer ip only if iptables setupIptables is true

## Checklist

- [v] Title of the PR starts with chart name (e.g. `[stable/mychartname]`)
- [v] I have read the [contribution instructions](https://github.com/deliveryhero/helm-charts#opening-a-pr), bumped chart version and regenerated the docs
- [v] Github actions are passing

### DIFF
--- a/stable/node-local-dns/Chart.yaml
+++ b/stable/node-local-dns/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: node-local-dns
-version: 1.1.6
+version: 1.2.0
 appVersion: 1.22.23
 maintainers:
 - name: gabrieladt
--- a/stable/node-local-dns/Chart.yaml
+++ b/stable/node-local-dns/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: node-local-dns
-version: 1.1.5
-appVersion: 1.22.20
+version: 1.1.6
+appVersion: 1.22.23
 maintainers:
 - name: gabrieladt
   email: no-reply@deliveryhero.com
--- a/stable/node-local-dns/Chart.yaml
+++ b/stable/node-local-dns/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: node-local-dns
-version: 1.2.0
+version: 2.0.0
 appVersion: 1.22.23
 maintainers:
 - name: gabrieladt
--- a/stable/node-local-dns/README.md
+++ b/stable/node-local-dns/README.md
@@ -52,7 +52,6 @@ helm install my-release deliveryhero/node-local-dns -f values.yaml
 | config.commProtocol | string | `"force_tcp"` |  |
 | config.dnsDomain | string | `"cluster.local"` |  |
 | config.dnsServer | string | `"172.20.0.10"` |  |
-| config.healthPort | int | `8080` |  |
 | config.localDns | string | `"169.254.20.25"` |  |
 | config.setupInterface | bool | `true` |  |
 | config.setupIptables | bool | `true` |  |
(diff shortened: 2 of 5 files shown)


STASiAN:
@max-rocket-internet please approve!

stable/node-local-dns/templates/daemonset.yaml
```diff
@@ -61,8 +61,13 @@ spec:
         resources: {{- toYaml . | nindent 10 }}
         {{- end }}
         args:
+        {{- if .Values.config.setupIptables }}
```
max-rocket-internet, line 64: Here you're essentially wrapping the existing args in a boolean that is currently `false` by default. So this change would adjust every user's args now to `-localip {{ .Values.config.localDns }}` unless they have already set `.Values.config.setupIptables=false`.

Do you that's desired?

max-rocket-internet:
I think you should also write a better description of your change and why 🙂

stable/node-local-dns/templates/daemonset.yaml
STASiAN: Thanks for fast reply.

My colleague @kkonnov mistakenly changed defaults in this [commit](https://github.com/deliveryhero/helm-charts/pull/503/commits/3b97be911095083e8e94e50d7a63e211639ca172).
For now I corrected defaults to return backward compatibility. 
JFYI: we are using specific configuration with cilium [local-policy-redirect](https://github.com/cilium/cilium/tree/main/examples/kubernetes-local-redirect) and don't iptables at all

STASiAN:
> I think you should also write a better description of your change and why 🙂

See my comments to your previous post

STASiAN:
Description has been updated

max-rocket-internet:
> JFYI: we are using specific configuration with cilium [local-policy-redirect](https://github.com/cilium/cilium/tree/main/examples/kubernetes-local-redirect) and don't iptables at all

OK cool but do you think changing the default behaviour of chart to suit Cilium for everyone is a good idea?

stable/node-local-dns/README.md
```diff
@@ -54,9 +54,9 @@ helm install my-release deliveryhero/node-local-dns -f values.yaml
 | config.dnsServer | string | `"172.20.0.10"` |  |
 | config.healthPort | int | `8080` |  |
 | config.localDns | string | `"169.254.20.25"` |  |
-| config.setupInterface | bool | `false` |  |
-| config.setupIptables | bool | `false` |  |
-| config.skipTeardown | bool | `true` |  |
```
max-rocket-internet, line 59: These are all arguments passed to the Daemonset, so changing these will change default behaviour for any user has not explicitly set them

STASiAN: The defaults was broken since this [commit](https://github.com/deliveryhero/helm-charts/pull/503/commits/3b97be911095083e8e94e50d7a63e211639ca172).

```node-cache --help```
``` 
  -setupinterface
    	indicates whether network interface should be setup (default true)
  -setupiptables
    	indicates whether iptables rules should be setup (default true)
  -skipteardown
    	indicates whether iptables rules should be torn down on exit
```

I just try to fix them ;)

**Comment by STASiAN** [comment]
@max-rocket-internet 
For now I have updated the minor version, because there have been many changes.
Is something else is required to be merged?

File: stable/node-local-dns/Chart.yaml [hunk 3]
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: node-local-dns
-version: 1.1.5
-appVersion: 1.22.20
+version: 1.2.0
```
max-rocket-internet, line 3: Since it's quite a lot changes and some of these might change the default behaviour of the chart, let's go to version 2.

```suggestion
version: 2.0.0
```

max-rocket-internet (approved):

max-rocket-internet:
Thanks @STASiAN

PR #534, opened 2023-10-17
[stable/node-local-dns]: restrict coredns binding

thomaschaaf:
## Description

Currently the coredns is bound very openly (0.0.0.0)

## Checklist


thomaschaaf:
I don't understand why the ci and my local machine in generating different docs. Do you have a hint for me?

max-rocket-internet:
Hey @thomaschaaf did you run the commands listed here from repo root? https://github.com/deliveryhero/helm-charts#running-ci-tests-locally

stable/node-local-dns/templates/configmap.yaml
```diff
@@ -19,7 +19,7 @@ data:
         }
         reload
         loop
-        bind 0.0.0.0
+        bind {{ .Values.config.localDns }} {{ .Values.config.dnsDomain }}
```
max-rocket-internet, line 22: Do either:
1. Use values there that generate the same output as before (`0.0.0.0`)
2. Bump the major chart version since this would be a breaking change for any users of this chart

Probably 1 is better IMO.

max-rocket-internet (changes requested):

stable/node-local-dns/templates/configmap.yaml
max-rocket-internet: > I don't think 0.0.0.0 should be the default

Why?

Also, now won't most users have to set `config.localDns` and `config.dnsDomain` to suit their environment? Or will these default values work in most cases?

thomaschaaf: When setting this up you need to check the configuration to see if the respective ips are unused. In our case the default is actually used and we switched to a different ip. See the note in the configuration part here: https://kubernetes.io/docs/tasks/administer-cluster/nodelocaldns/#configuration

The manifest which kubernetes provides doesn't listen on all interfaces either but is more restrictive. As this is running in the hostNetwork there could be interfaces which should not expose a dns server. Maybe an interface which is connected with a customers vpn or a different vpc. https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml#L77

max-rocket-internet: We should definitely make it configurable, which it presently is not, but I'm still trying to decide if this is a good idea to merge or not 🤔 As far as I understand, this default value works for everyone.

> coredns is bound very openly (0.0.0.0)
> I don't think 0.0.0.0 should be the default

Why does this matter? What's the rationale for for removing this for a DNS caching service?

PR #539, opened 2023-11-03
[stable/field-exporter]: field-exporter

vzholudev:
## Description

  A chart to install [field-exporter](https://github.com/deliveryhero/field-exporter). This controller is used to fill the gap in [k8s-config-connector](https://github.com/GoogleCloudPlatform/k8s-config-connector) for exporting value from Config Connector managed resources into Secrets and ConfigMaps.

## Checklist

### DIFF
--- a/stable/field-exporter/templates/deployment.yaml
+++ b/stable/field-exporter/templates/deployment.yaml
@@ -1,0 +1,56 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "field-exporter.fullname" . }}-controller-manager
+  labels:
+    app.kubernetes.io/component: manager
+    app.kubernetes.io/created-by: field-exporter
+    app.kubernetes.io/part-of: field-exporter
+    control-plane: controller-manager
+  {{- include "field-exporter.labels" . | nindent 4 }}
+spec:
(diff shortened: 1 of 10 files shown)


stable/field-exporter/templates/deployment.yaml
```diff
@@ -0,0 +1,56 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "field-exporter.fullname" . }}-controller-manager
+  labels:
+    app.kubernetes.io/component: manager
+    app.kubernetes.io/created-by: field-exporter
+    app.kubernetes.io/part-of: field-exporter
+    control-plane: controller-manager
```
max-rocket-internet, line 9: What is this label for?

vzholudev: @arjunrn or @mtahaahmed can you answer this question?

mtahaahmed: This helm chart was converted from kustomize. I wanted to change in the helm chart after I see it running in the cluster.

mtahaahmed: we can probably remove it for now

PR #541, opened 2023-11-06
[stable/k8s-event-logger] Added option to change container name

shehbazk:
Made changes in deployment.yaml and values.yaml to add an option to change container name. Also set the default to "k8s-event-logger".

### DIFF
--- a/stable/k8s-event-logger/templates/deployment.yaml
+++ b/stable/k8s-event-logger/templates/deployment.yaml
@@ -37,7 +37,7 @@ spec:
 {{- end }}
       serviceAccountName: {{ include "k8s-event-logger.fullname" . }}
       containers:
-        - name: {{ .Values.image.containerName}}
+        - name: {{ .Values.containerName}}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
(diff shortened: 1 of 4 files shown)


stable/k8s-event-logger/values.yaml
```diff
@@ -1,4 +1,5 @@
 image:
+  containerName: k8s-event-logger
```
max-rocket-internet, line 2: It's not part of the image so I would not put it under here:

```suggestion
containerName: k8s-event-logger

image:
```

stable/k8s-event-logger/templates/deployment.yaml
```diff
@@ -37,7 +37,7 @@ spec:
 {{- end }}
       serviceAccountName: {{ include "k8s-event-logger.fullname" . }}
       containers:
-        - name: app
+        - name: {{ .Values.image.containerName}}
```
max-rocket-internet, line 40:
```suggestion
        - name: {{ .Values.containerName}}
```

max-rocket-internet (changes requested):
You need read the contribution notes, there is details on what is required in a PR (e.g. regenerate the chart README, bump version etc)

max-rocket-internet (approved):
Thanks @shehbazk 💜

PR #557, opened 2024-01-08
[stable/cachet] DB_DRIVER defined twice

fcecagno:
## Description

`DB_DRIVER` is being defined twice in the configmap, which makes helm fail with:
```
Helm upgrade failed: error while running post render on files: map[string]interface {}(nil): yaml: unmarshal errors:
  line 28: mapping key "DB_DRIVER" already defined at line 19
```
Removed one of the properties and left the `env.public.DB_DRIVER`, which had the correct default value `pgsql`.

Version 1.3.3 only works if `env.public.DB_DRIVER` is defined as empty, so it's ignored and not set in the configmap.

## Checklist

### DIFF
--- a/stable/cachet/Chart.yaml
+++ b/stable/cachet/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "2.3.15"
 description: The open source status page system
 name: cachet
-version: 1.3.3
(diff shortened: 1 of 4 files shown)


max-rocket-internet (approved):
Cool, thanks @fcecagno 💗

PR #576, opened 2024-04-08
[stable/NONE]: update README, make helm-docs instructions robust

stxm:
## Description

 `${PWD}` implies that the command could be run from any directory within the repo. This is not true since the path to `./ci/README.md.gotmpl` is relative. 

## Checklist


max-rocket-internet (approved):
At the top of this section it says:

> All commands to be run from the root of this repo

But cool, if it helps someone then all good!

PR #586, opened 2024-06-19
[stable/node-local-dns] add prefetch options to configuration

andrii29:
## Description
add ability to enable [prefetch](https://coredns.io/plugins/cache/)

prefetch will prefetch popular items when they are about to be expunged from the cache. Popular means AMOUNT queries have been seen with no gaps of DURATION or more between them. DURATION defaults to 1m. Prefetching will happen when the TTL drops below PERCENTAGE, which defaults to 10%, or latest 1 second before TTL expiration. Values should be in the range [10%, 90%]. Note the percent sign is mandatory. PERCENTAGE is treated as an int.
## Checklist

### DIFF
--- a/stable/node-local-dns/Chart.yaml
+++ b/stable/node-local-dns/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: node-local-dns
-version: 2.0.11
+version: 2.0.12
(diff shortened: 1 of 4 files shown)


max-rocket-internet (approved):
Interesting feature, thanks @andrii29 💜

PR #588, opened 2024-06-30
sync labels between deployment and pod

BlueHorn07:
## Checklist

### DIFF
--- a/stable/aws-s3-proxy/templates/deployment.yaml
+++ b/stable/aws-s3-proxy/templates/deployment.yaml
@@ -8,13 +8,11 @@ spec:
   replicas: {{ .Values.replicaCount }}
   selector:
     matchLabels:
-      app.kubernetes.io/name: {{ include "aws-s3-proxy.name" . }}
-      app.kubernetes.io/instance: {{ .Release.Name }}
+{{ include "aws-s3-proxy.labels" . | indent 6 }}
   template:
     metadata:
       labels:
-        app.kubernetes.io/name: {{ include "aws-s3-proxy.name" . }}
-        app.kubernetes.io/instance: {{ .Release.Name }}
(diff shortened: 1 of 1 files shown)


stable/aws-s3-proxy/templates/deployment.yaml
```diff
@@ -8,13 +8,11 @@ spec:
   replicas: {{ .Values.replicaCount }}
   selector:
     matchLabels:
-      app.kubernetes.io/name: {{ include "aws-s3-proxy.name" . }}
-      app.kubernetes.io/instance: {{ .Release.Name }}
+{{ include "aws-s3-proxy.labels" . | indent 6 }}
```
max-rocket-internet, line 11: I don't think this part should change as selector labels are immutable and shouldn't be treated like the other labels 🙂

max-rocket-internet (changes requested):

stable/aws-s3-proxy/templates/deployment.yaml
BlueHorn07: Hi there, thanks for your comment!

Are you suggesting that the changes to `matchLabels` might cause breaking changes? If that’s the case, I’ll revert the changes to `matchLabels`. 🙏

max-rocket-internet: Exactly, they are immutable, you can't change them without deleting and recreating the resource. So I would just not change this part.

BlueHorn07: I’ve reverted it. Thanks for pointing it out and for your kind review! 😄

max-rocket-internet (approved):
Nice, thanks @BlueHorn07 💖

PR #604, opened 2024-09-20
[stable/node-local-dns] fix: prevent duplicate metrics

### LINKED ISSUE #600: [node-local-dns] duplicated metrics
By default the coredns metrics get scraped twice, since both the service and the pods have the annotations

prometheus.io/port: "9253"
prometheus.io/scrape: "true"

Since the service has ClusterIP: None each pod will be scraped. So if you issue a PromQL query like sum(rate(coredns_dns_requests_total{k8s_app=~"node-local-dns"}[5m])) by (proto) the results will be double the "actual" value. Sure, this can be fixed with adding for example job = "kubernetes-service-endpoints" to the query, but that complicates the query unnecessarily and you need to understand the problem to figure it out.

I suggest removing the prometheus annotations from the pod. Removing them from the service would also work, but since you could add them back yourself to the pod using podAnnotations I think it's better to remove them from there. In the service you would still be able to control them using prometheusScraping.enabled.

msvticket:
Fixes #600

## Description

Fixes #600

## Checklist


max-rocket-internet (approved):
Thanks @msvticket 💚

Issue #607, opened 2024-10-02
Move to a more reliable chart hosting solution


max-rocket-internet:
We have periodic problems with chartmuseum running on k8s, e.g:
https://github.com/deliveryhero/helm-charts/issues/606
https://github.com/deliveryhero/helm-charts/issues/554

GHCR has [been suggested](https://github.com/deliveryhero/helm-charts/issues/591) as a good option.

pierluigilenoci:
@max-rocket-internet, you have already some PR ready to fix this:
https://github.com/max-rocket-internet/k8s-event-logger/pull/45

shubhamv1436:
@max-rocket-internet Its still happening. Anything you can do to continuously monitor and automate the fix.

```
helm repo add deliveryhero https://charts.deliveryhero.io/
Error: looks like "https://charts.deliveryhero.io/" is not a valid chart repository or cannot be reached: failed to fetch https://charts.deliveryhero.io/index.yaml : 504 Gateway Timeout
```

msvticket:
Considering that you have Cloudflare configured as CDN in front of the Chartmuseum a quick work around (while working on a long term solution) should be to configure Cloudflare to serve content even though it is is stale (i.e. isn't valid according to cache headers).

bunnybilou:
You might want to consider adding caching on Cloudflare to reduce the need for frequent origin fetches.

I checked the Argo Helm repository, and they have a max-age of 600 seconds configured for caching: https://argoproj.github.io/argo-helm/index.yaml

max-rocket-internet:
I would be grateful if everyone could take a look at [this PR](https://github.com/deliveryhero/helm-charts/pull/612) 😄

msvticket:
Unfortunately it doesn't work properly; there seem to be some kind of permission problem:

```
$ helm template oci://ghcr.io/deliveryhero/helm-charts/node-local-dns
Error: GET "https://ghcr.io/v2/deliveryhero/helm-charts/node-local-dns/tags/list": GET "https://ghcr.io/token?scope=repository%3Adeliveryhero%2Fhelm-charts%2Fnode-local-dns%3Apull&service=ghcr.io": unexpected status code 401: unauthorized: authentication required
```

It is not a general problem with ghcr. An example of another repository that works is oci://ghcr.io/bank-vaults/helm-charts/vault-operator

msvticket:
Another problem is that the instructions are incomplete:

```
[...]

Issue #611, opened 2024-10-04
[stable/k8s-event-logger] - latest chart not avaiable


jaimin-shah-fs:
There was an update to the stable chart for k8s-event-logger yesterday, but when accessing `https://chart.deliveryhero.io/index.yaml` the latest version still shows 1.1.7 instead of 1.1.8. Is there a delay in chartmuseum before the chart is available publically?

max-rocket-internet:
Yes, there's a temporary cache in place, see https://github.com/deliveryhero/helm-charts/issues/607, I'll hopefully have a fix today 🙏

max-rocket-internet:
@jaimin-shah-fs can you please check with the new OCI hosted chart: https://github.com/deliveryhero/helm-charts/tree/master/stable/k8s-event-logger#how-to-install-this-chart

PR #612, opened 2024-10-07
Move chart hosting to GHCR

max-rocket-internet:
Here I am replacing the old [chartmuseum](https://chartmuseum.com/) based hosting with [OCI](https://helm.sh/docs/topics/registries/) on [Github Container Registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry)

Resolves https://github.com/deliveryhero/helm-charts/issues/607 & https://github.com/deliveryhero/helm-charts/issues/611

- Replacing old github action with a new one that uses helm directly to `push` the packaged charts to GHCI
- Updating template used to generate all helm chart READMEs
- Update all chart READMEs with new templated text
- Updating main README
- Removing old gif that showed adding the repo


max-rocket-internet:
In this special case I'm gonna ignore those 2x CI tests as they are not applicable here.

max-rocket-internet:
Maybe after merging I could go through the git history and run `helm push` all old chart versions also?

mKeRix (approved):

PR #631, opened 2024-11-16
[stable/locust]: Updates the service and ingress to accept more port configurability

byronmansfield:
## Description

Updates the service and ingress to accept more port configurability

## Checklist

### DIFF
--- a/stable/locust/README.md
+++ b/stable/locust/README.md
@@ -106,25 +106,16 @@ helm install my-release oci://ghcr.io/deliveryhero/helm-charts/locust -f values.
 | master.envs_include_default | bool | `true` | Whether to include default environment variables |
 | master.extraPorts | string | `nil` |  |
 | master.image | string | `""` | A custom docker image including tag |
-| master.livenessProbe.enabled | bool | `false` |  |
-| master.livenessProbe.failureThreshold | int | `2` |  |
-| master.livenessProbe.initialDelaySeconds | int | `60` |  |
-| master.livenessProbe.path | string | `"/"` |  |
-| master.livenessProbe.periodSeconds | int | `30` |  |
-| master.livenessProbe.port | int | `8089` |  |
-| master.livenessProbe.scheme | string | `"HTTP"` |  |
-| master.livenessProbe.successThreshold | int | `1` |  |
-| master.livenessProbe.timeoutSeconds | int | `30` |  |
+| master.livenessProbe | object | `{}` |  |
 | master.logLevel | string | `"INFO"` | Log level. Can be INFO or DEBUG |
 | master.nodeSelector | object | `{}` | Overwrites nodeSelector from global |
 | master.pdb.enabled | bool | `false` | Whether to create a PodDisruptionBudget for the master pod |
-| master.readinessProbe.enabled | bool | `true` |  |
 | master.readinessProbe.failureThreshold | int | `2` |  |
+| master.readinessProbe.httpGet.path | string | `"/"` |  |
+| master.readinessProbe.httpGet.port | int | `8089` |  |
+| master.readinessProbe.httpGet.scheme | string | `"HTTP"` |  |
 | master.readinessProbe.initialDelaySeconds | int | `5` |  |
-| master.readinessProbe.path | string | `"/"` |  |
 | master.readinessProbe.periodSeconds | int | `30` |  |
-| master.readinessProbe.port | int | `8089` |  |
-| master.readinessProbe.scheme | string | `"HTTP"` |  |
 | master.readinessProbe.successThreshold | int | `1` |  |
 | master.readinessProbe.timeoutSeconds | int | `30` |  |
 | master.replicas | int | `1` | Should be set to either 0 or 1. |
--- a/stable/locust/README.md
+++ b/stable/locust/README.md
@@ -122,6 +122,7 @@ helm install my-release oci://ghcr.io/deliveryhero/helm-charts/locust -f values.
 | master.resources | object | `{}` | resources for the locust master |
 | master.restartPolicy | string | `"Always"` | master pod's restartPolicy. Can be Always, OnFailure, or Never. |
 | master.serviceAccountAnnotations | object | `{}` |  |
+| master.startupProbe | object | `{}` |  |
 | master.strategy.type | string | `"RollingUpdate"` |  |
 | master.tolerations | list | `[]` | Overwrites tolerations from global |
 | nameOverride | string | `""` |  |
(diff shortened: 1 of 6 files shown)


byronmansfield:
@max-rocket-internet I created this PR out of some needs from working with this chart in GKE. It required a few service/ingress specific settings that this would help me out greatly for container native loadbalancing in GKE. The readiness/liveness changes apparently control the health checks. I also have the need to expose it under a different port. I hope these changes are acceptable. Since it was a little larger of a change, I hope I made the right judgment call on versioning. I tried to do my part in local testing, though I don't have a personal GKE or AWS account to actually run it through a more enterprise setting. Let me know what you think.

max-rocket-internet:
Thanks for the effort and explanation @byronmansfield!

I think it's mergeable as is but raised one point [here](https://github.com/deliveryhero/helm-charts/pull/631#discussion_r1846542828)

stable/locust/README.md
```diff
@@ -104,10 +104,29 @@ helm install my-release oci://ghcr.io/deliveryhero/helm-charts/locust -f values.
...
 | master.image | string | `""` | A custom docker image including tag |
+| master.livenessProbe.enabled | bool | `false` |  |
+| master.livenessProbe.failureThreshold | int | `2` |  |
+| master.livenessProbe.initialDelaySeconds | int | `60` |  |
+| master.livenessProbe.path | string | `"/"` |  |
+| master.livenessProbe.periodSeconds | int | `30` |  |
+| master.livenessProbe.port | int | `8089` |  |
+| master.livenessProbe.scheme | string | `"HTTP"` |  |
+| master.livenessProbe.successThreshold | int | `1` |  |
+| master.livenessProbe.timeoutSeconds | int | `30` |  |
```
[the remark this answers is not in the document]
byronmansfield: Excellent! Yes, the thought had also crossed my mind to do it this way. I've seen it done this way in other public charts too. I'll refactor this later to fit your suggestion. Thanks for your time in reviewing this and making solid suggestions. I know doing this for all the charts you have in here probably is no trivial task.

One last question before I make this change later. Do you want me to additionally make `master.readinessProbe` to also be one object? I'm ok with it either way. But I'd say keep them similar.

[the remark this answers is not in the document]
max-rocket-internet: > Do you want me to additionally make master.readinessProbe to also be one object? 

I would say yes, for consistency 🙂

max-rocket-internet:
Current CI error:

```
Error: INSTALLATION FAILED: Deployment.apps "locust-6sejfzxok0-master" is invalid: spec.template.spec.containers[0].livenessProbe: Required value: must specify a handler type
```

🤔

byronmansfield:
@max-rocket-internet thanks. Either I can't see the CI output because I'm not part of the group/org which is understandable. Or I've been in gitlab/jenkins ci space too long I can't seem to figure out where to look. Either way. Here is what I can give you from my knowledge (don't claim to be a k8s ninja) and local testing. Plus some good ol fashion googling.

I've been using `helm template locust -n locust -f test.values.yaml --version 0.31.9 --debug . | kubeconform -strict -verbose -summary` locally with different variations of the `test.values.yaml`. With an empty livenssProbe, and with some basic settings without the exec or httpGet. And then some with exec and httpGet. And `kubeconform` comes back ok. I also just output to a local `debug.yaml` without pipeing it to `kubeconform`. To do a manual inspection. All seems to be ok from my local testing. Maybe the CI values are doing a configuration I haven't tested yet.

Many times when working on helm charts. I lean on large feature full complex projects such as [vault's](https://github.com/hashicorp/vault-helm/tree/main) or [jenkin's](https://github.com/jenkinsci/helm-charts/tree/main/charts/jenkins) helm charts, just to see how they are handling it. They have years of development behind them and people doing all kinds of wild configurations. I cannot see anything immediately from their variations of how they handle setting up their probes. Jenkins does one more like this style.

Is it possible to share a bit more as to what the CI jobs are testing with so that I can try to troubleshoot? I tried looking at the [ci/ct.yaml](https://github.com/deliveryhero/helm-charts/blob/master/ci/ct.yaml) (I'm assuming this is the failing test) although I'm confused as to what it actual does.

Sorry I can't be more help.

max-rocket-internet:
> Either I can't see the CI output because I'm not part of the group/org which is understandable

Any logged in user can see it: https://github.com/deliveryhero/helm-charts/actions/runs/11905140650/job/33175111907#step:9:23

> Oh, last thing. Once we figure this out. How would you feel about me adding in a startupProbe just to cover all basis.

Do it 🚀

> Maybe the CI values are doing a configuration I haven't tested yet.

I'll take a look 🕒

stable/locust/templates/master-deployment.yaml
```diff
@@ -141,24 +144,13 @@ spec:
...
-            port: 8089
-{{- if .Values.master.auth.enabled }}
-            httpHeaders:
-                - name: Authorization
-                  value: Basic {{ printf "%s:%s" .Values.master.auth.username .Values.master.auth.password | b64enc }}
-{{- end }}
-{{- end }}
+          {{- toYaml .Values.master.readinessProbe | nindent 12 }}
+        livenessProbe:
+          {{- toYaml .Values.master.livenessProbe | nindent 12 }}
```
max-rocket-internet, line 153: For whatever reason, you gotta do this:

```suggestion
        {{- if .Values.master.livenessProbe }}
        livenessProbe:
          {{- toYaml .Values.master.livenessProbe | nindent 12 }}
        {{- end }}
```

max-rocket-internet:
> I've been using helm template ..... And kubeconform comes back ok

Yes, quite annoying this error is not catchable until it gets sent to a real k8s API. Well, I guess that's why we have the "chart testing" CI test. I've suggested change on your PR that should fix it 🙂

stable/locust/templates/master-deployment.yaml
byronmansfield: Awesome. Great catch. Thank you!

stable/locust/README.md
```diff
@@ -104,14 +104,25 @@ helm install my-release oci://ghcr.io/deliveryhero/helm-charts/locust -f values.
 | master.deploymentAnnotations | object | `{}` | Annotations on the deployment for master |
 | master.environment | object | `{}` | environment variables for the master |
 | master.envs_include_default | bool | `true` | Whether to include default environment variables |
+| master.extraPorts | string | `nil` |  |
 | master.image | string | `""` | A custom docker image including tag |
+| master.livenessProbe | object | `{}` |  |
 | master.logLevel | string | `"INFO"` | Log level. Can be INFO or DEBUG |
 | master.nodeSelector | object | `{}` | Overwrites nodeSelector from global |
 | master.pdb.enabled | bool | `false` | Whether to create a PodDisruptionBudget for the master pod |
+| master.readinessProbe.failureThreshold | int | `2` |  |
```
max-rocket-internet, line 113: I think we do the same here, like `livenessProbe`, yeah?

max-rocket-internet: Ahh no, this copies the existing values, nice

max-rocket-internet (approved):
OK nice, thanks @byronmansfield

PR #634, opened 2024-11-21
[stable/locust]: Updates configmap for locustfile and lib to not be wrapped in explictstring "example"

byronmansfield:
## Description

Updates configmap for locustfile and lib to not be wrapped in explictstring "example"

## Checklist

### DIFF
--- a/stable/locust/Chart.yaml
+++ b/stable/locust/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: locust
-version: "0.32.0"
+version: "0.32.1"
 appVersion: 2.32.2
 home: https://github.com/locustio/locust
 icon: https://locust.io/static/img/logo.png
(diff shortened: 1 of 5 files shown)


byronmansfield:
Hey @max-rocket-internet I hope you are not getting sick of me yet. :laughing:

I made this one because I have been using helmfile and made my own locustfile `ConfigMap` I deployed from a local chart in the same repo. But I always had to roll the locust pods in order for them to get pick up for the locustfile change. Which was annoying. Then I tried to add the checksum annotation, and after re-reading the helm chart, I realized, that only works for the same helm deployment. You can't do that natively in kubernetes without something like reloader.

The current helm chart locustfile (and lib) CM's only will deploy if they are the default "example". I hope it's ok we just default it to that in the values.yaml and still deploy the same CM if someone overrides those values.

I believe the README.md explicitly calls out you have to create your own CM. If you would prefer I update the README.md to reflect these changes. I am happy to add that. Thoughts?

max-rocket-internet (approved):
Nice, thanks @byronmansfield 💜

PR #636, opened 2024-11-27
[stable/locust]: Updates locust chart configmap files glog statement

byronmansfield:
## Description

Updates the locustfile's configmap Files Glob syntax. Seems to be a small syntax change for helm 3 maybe.

## Checklist

### DIFF
--- a/stable/locust/Chart.yaml
+++ b/stable/locust/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: locust
-version: "0.32.1"
+version: "0.32.2"
 appVersion: 2.32.2
 home: https://github.com/locustio/locust
 icon: https://locust.io/static/img/logo.png
--- a/stable/locust/README.md
+++ b/stable/locust/README.md
@@ -1,6 +1,6 @@
 # locust
 
-![Version: 0.32.1](https://img.shields.io/badge/Version-0.32.1-informational?style=flat-square) ![AppVersion: 2.32.2](https://img.shields.io/badge/AppVersion-2.32.2-informational?style=flat-square)
+![Version: 0.32.2](https://img.shields.io/badge/Version-0.32.2-informational?style=flat-square) ![AppVersion: 2.32.2](https://img.shields.io/badge/AppVersion-2.32.2-informational?style=flat-square)
 
 A chart to install Locust, a scalable load testing tool written in Python.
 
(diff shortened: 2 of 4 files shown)


byronmansfield:
Hey @max-rocket-internet. So with my last update I was only testing it with `helm template locust -n locust -f values.yaml testing.values.yaml --version 0.32.1 --debug . > debug.yaml` and the locustfile configmap rendered the data just fine, as expected. With the full python file in there. Now in my `testing.values.yaml` I was testing with trying to use `stable/locust/locustfiles/foo/main.py`. And had updated this in the values file according to `loadtest.name: foo` and `loadtest.locust_locustfile: main.py`.

However, today I went to actually deploy that latest version (0.32.1) at work. And it did not work. I spent some time troubleshooting it. At work I cannot run `helm template` because that is only for local charts. I got home and troubleshot it locally with `helm upgrade -n locust --dry-run -n locust -f values.yaml -f test.values.yaml --version 0.32.1 oci://ghcr.io/deliveryhero/helm-charts/locust > debug.yaml` and sure enough, it was still not rendering like `helm template ...` does. Odd. 

So I eventually got it working like this and instead of using the remote one like my `helm upgrade` above. I just switched it out for `.` (locally). It seems to work this way.

Let me know your thoughts and if it needs more testing. Sorry I didn't catch this before with `helm template`. It's odd to me (and I'm not that experienced with contributing to helm charts) that helm template and helm upgrade --dry-run render two different things for the locustfile configmap.

I hope all my comment made sense.

max-rocket-internet:
Hey @byronmansfield, I don't have time to do thorough reviews of all these PRs...

The configmap created by `stable/locust/templates/configmap-locust-locustfile.yaml` is just an example to show how it works. i.e. the format of the configmap and how it's mounted inside the locust pod. This configmap cannot be used if you are using your own locustfile (as you should be) because helm can only access files within the chart itself. This is stated already in the readme:

> By default it will install using an example locustfile and lib from [stable/locust/locustfiles/example](https://github.com/deliveryhero/helm-charts/tree/master/stable/locust/locustfiles/example). When you want to provide your own locustfile, you will need to create 2 configmaps using the structure from that example

So I think just revert `stable/locust/templates/configmap-locust-locustfile.yaml` back to the version before your https://github.com/deliveryhero/helm-charts/pull/634 PR.

Otherwise this configmap is created for no reason in most cases, right?

byronmansfield:
I hope just adding back the old conditionals in this PR is ok. And thanks for the `--set-file` advice. That is a pretty cool feature. I actually never knew about this. That might actually do the trick.

max-rocket-internet (approved):
All good, thanks @byronmansfield 💖

PR #641, opened 2024-12-16
[stable/locust] Add support for git sync

AlexisBRENON:
## Description

Add Git Sync support to locust chart to easily deploy multiple locustfiles.

## Checklist

### DIFF
--- a/stable/locust/templates/worker-deployment.yaml
+++ b/stable/locust/templates/worker-deployment.yaml
@@ -45,11 +45,18 @@ spec:
 {{- if .Values.hostAliases }}
       hostAliases: {{- .Values.hostAliases | toYaml | nindent 8 }}
 {{- end }}
+      initContainers:
+        {{- if .Values.locustfiles.gitSync.enabled }}
+          {{- include "git_sync_container" (dict "Values" .Values "is_init" "true" "Template" .Template) | nindent 8 }}
+        {{- end }}
       containers:
+      {{- if .Values.locustfiles.gitSync.enabled }}
+        {{- include "git_sync_container" . | nindent 8 }}
+      {{- end }}
       - name: {{ .Chart.Name }}
         securityContext:
           {{- toYaml .Values.securityContext | nindent 12 }}
-        image: "{{ if .Values.worker.image }}{{ .Values.worker.image }}{{ else }}{{ .Values.image.repository }}:{{ .Values.image.tag }}{{ end }}"
+        image: {{ template "locust.worker_image" }}
 {{- with .Values.worker.command }}
         command:
         {{- toYaml . | nindent 8 }}
@@ -70,6 +77,9 @@ spec:
           {{- if .Values.loadtest.locust_lib_configmap }}
           - name: lib
             mountPath: "{{ .Values.loadtest.locust_locustfile_path }}/lib"
+          {{- end }}
+          {{- if .Values.locustfiles.gitSync.enabled }}
+            {{- include "locust.locustfiles_mount" . | nindent 12 }}
           {{- end }}
           - name: config
             mountPath: /config
@@ -132,6 +142,10 @@ spec:
         - name: locustfile
           configMap:
             name: {{ .Values.loadtest.locust_locustfile_configmap }}
+        {{- end }}
+        {{- if .Values.dags.gitSync.enabled }}
+        - name: locustfiles
+          emptyDir:
         {{- end }}
         - name: config
           configMap:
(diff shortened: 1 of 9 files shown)


max-rocket-internet:
Sorry @AlexisBRENON but this is a huge increase in complexity, with no description of why and no docs for how a chart user might make use of this feature.

max-rocket-internet:
How about a more generic approach that moves all this complexity to the values provided to the chart (and hence the person wanting to do this):
- Add a `.Values.initContainers` option
- Add a `.Values.volumes` option

This would be very simple and also, AFAIU, would allow you to achieve the same thing?

AlexisBRENON:
Thanks for the feedback. I agree, I didn't take time to document the new features, maybe I opened the PR a little bit early (moreover, the template is not rendering for the moment...).

Using a more generic approach report the complexity on the user of the chart. Moreover, it also requires a `.Values.sidecars` to allow for the git sync sidecar... Once documented, you will see that the use is quiet straight-forward, allowing the user to use any Git repository as the source of multiple locustfiles (as well as libraries, requirements file, etc.).
I tried to not break the current compatibility.

For information, I took inspiration from the Apache Airflow chart which can also relies on Git Sync to provide DAG files to the different components.

Are you totally againt such solution (in which case I can close the PR) or would you let me improve it a little bit (at least adding the documentation) (in which case I will ping you back when it's done) ? Or I can close the current PR and open a new one when I'm ready.

max-rocket-internet:
I'm on holiday right now, I'll revisit it when I get back in Jan, but in principle if there's a bunch of people that want this feature then all good, of course we can fix the issues and merge. But it looks like quite an advanced use case and from what I've seen so far, from the locust repo issues and the locust Slack org, even just using the helm chart as it is now can be a challenge.

If anyone else has strong opinions on this then feel free to chime in 😃

Happy holidays @AlexisBRENON 🎄

AlexisBRENON:
Hi.

I fixed the rendering issue and tried to add more documentation.

There is still some stuffs that I copied from the Airflow chart, without exactly knowing what it is used for (lifecycle hooks and security context). I could drop them if you want to keep the code more smaller, but I suspect that it may be of use for certain kind of deployment.

Finally, the refactor to allow different images between master and worker may be in another PR, what do you think ? But in this case, where should I put the git Sync image configuration (in `gitSync.image` or `gitSyncImage`?). I mainly did it to be consistent between master/worker/gitSync image configuration.

stable/locust/values.yaml
```diff
@@ -36,11 +37,105 @@ loadtest:
...
+    tag: ~
+    pullPolicy: IfNotPresent
+
+  gitSync:
+    # images.gitSync.repository -- image used by gitSync container
+    repository: registry.k8s.io/git-sync/git-sync
+    # images.gitSync.tag -- tag used by gitSync container
+    tag: v4.1.0
+    pullPolicy: IfNotPresent
+
```
max-rocket-internet, line 137: Why do we need all this complexity about images? Is it related to git sync?

stable/locust/values.yaml
```diff
@@ -36,11 +37,105 @@ loadtest:
   # loadtest.locustCmd -- The command to run Locust
   locustCmd: "/opt/venv/bin/locust"
 
+locustfiles:
+  # locustfiles.mountPath -- the path of the locustfiles (without trailing backslash)
+  mountPath: "/mnt/locust"
+  # locustfiles.requirements -- Path to a file containing requirements to install
+  requirements: ~
+  gitSync:
```
max-rocket-internet, line 45: I think you can simply remove a lot of they keys in here to make it simpler, e.g. containerName, uid, securityContext, securityContexts, containerLifecycleHooks, extraVolumeMounts etc.

Do we really need to set all these just to get git-sync working? I would say just leave all these as k8s defaults if not

stable/locust/templates/worker-deployment.yaml
```diff
@@ -45,11 +45,18 @@ spec:
 {{- if .Values.hostAliases }}
       hostAliases: {{- .Values.hostAliases | toYaml | nindent 8 }}
 {{- end }}
+      initContainers:
+        {{- if .Values.locustfiles.gitSync.enabled }}
```
max-rocket-internet, line 49: Does this not leave a key with no value if gitSync is not enabled?

max-rocket-internet:
It remains way too complicated IMO. I would say either make it much simpler or just create a separate chart, e.g. `locust-gitsync`. In this separate chart you can also remove loads of things, like the default configmaps etc.

max-rocket-internet:
Yeah now I think about it, let's do it as a separate chart and see how it evolves form there, how popular it is, how many contributions it gets etc 😃

PR #690, opened 2025-04-17
[stable/locust] Support worker.ports

RenaultAI:
## Description

My use case is for exposing a prom_http / prometheus port. The master already supports `extraPorts`.

## Checklist

### DIFF
--- a/stable/locust/Chart.yaml
+++ b/stable/locust/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: locust
-version: 0.32.5
+version: 0.32.6
 appVersion: 2.32.2
 home: https://github.com/locustio/locust
 icon: https://locust.io/static/img/logo.png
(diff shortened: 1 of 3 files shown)


max-rocket-internet (approved):
Cool, thanks @RenaultAI 💜

RenaultAI:
> Cool, thanks @RenaultAI 💜

Thanks! Unfortunately I think I needed to make the `ports` sectional conditional according to the error reported in https://github.com/deliveryhero/helm-charts/actions/runs/14524322050

`Error: INSTALLATION FAILED: Deployment in version "v1" cannot be handled as a Deployment: json: cannot unmarshal object into Go struct field Container.spec.template.spec.containers.ports of type []v1.ContainerPort`

I'm now able to run `ct install --config ci/ct.yaml` on my local kubernetes cluster. Previously I only ran the `ct lint --charts stable/locust` command which didn't error.

Issue #694, opened 2025-05-05
node local dns - Prometheus scraping is not working


SadmiB:
When using `Prometheus-operator ` enabling prometheus scarping with annotations is not supported, the way to go in this case is using `ServiceMonitor` but the node local dns chart hard codes the namespace of the service monitor which expects both prometheus and the pod to monitor in `kube-system` which is not the always the case.

The solutions should be: namespaces passed as values that we could customize.

I can contribute this change.

max-rocket-internet:
> I can contribute this change.

Do it 😃

PR #695, opened 2025-05-10
[stable/node-local-dns] Fix prometheus service monitor

SadmiB:
## Description

This is a small change that makes passing a custom namespace to the service monitor of node local dns possible;
This will be helpful in some situations where prometheus is not installed on kube-system namespace.

#fixes https://github.com/deliveryhero/helm-charts/issues/694

## Checklist


stable/node-local-dns/templates/servicemonitor.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   name: {{ include "node-local-dns.fullname" . }}
-  namespace: kube-system
+  namespace: {{ .Values.serviceMonitor.prometheusNamespace}}
```
max-rocket-internet, line 6: Is it typical to create this `ServiceMonitor` resource in a different namespace to the rest of the resources? If not, then I would say we could just make a single value for the namespace of all resources? Or alternatively remove `namespace` from all resources and rely on `--namespace` in Helm?

max-rocket-internet: OK, makes sense, thanks

SadmiB:
@max-rocket-internet not sure why the ci job check is complaining even if the readme is updated

stable/node-local-dns/README.md
```diff
@@ -12,37 +12,11 @@ Further documentation is [here](https://kubernetes.io/docs/tasks/administer-clus
 
 This helm chart works for both kube-proxy setups (iptables or ipvs).
 
-## How to install this chart
```
max-rocket-internet, line 15: This stuff should not be removed. Are you running `docker run --rm -v "$(git rev-parse --show-toplevel):/helm-docs" jnorwood/helm-docs:v1.11.3 --template-files ./ci/README.md.gotmpl`?

max-rocket-internet: Can you try running it from repo root @SadmiB ?

SadmiB: Nice, regenerated from root.

max-rocket-internet (approved):

PR #710, opened 2025-07-30
[stable/node-local-dns] add `port` option to config

nerdroychan:
## Description

When custom config in node-local-dns is used, users may run the node-local-dns on a non-53 port. One use case is running node-local-dns on host network and using cilium to redirect traffic. In that case, running the node cache in a non-standard port can avoid potential conflict with any programs running on the host using port 53.

This commit adds a `port` to chart values. It defaults to 53 if not overridden. But providing this options gives users flexibility to run the service on a different port.

## Checklist

### DIFF
--- a/stable/node-local-dns/Chart.yaml
+++ b/stable/node-local-dns/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: node-local-dns
-version: 2.1.8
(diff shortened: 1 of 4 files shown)


max-rocket-internet (approved):
Thanks @nerdroychan 💜

PR #787, opened 2026-04-03
[stable/node-problem-detector] Add initContainers support

aslafy-z:
## Description

Add initContainers support to the node-problem-detector DaemonSet, following the same pattern as the existing extraContainers field. Users can specify a list of init containers in their values to run before the main node-problem-detector container starts.

## Checklist

### DIFF
--- a/stable/node-problem-detector/Chart.yaml
+++ b/stable/node-problem-detector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: node-problem-detector
-version: 2.4.0
+version: 2.4.1
 appVersion: v1.35.1
 home: https://github.com/kubernetes/node-problem-detector
 description: |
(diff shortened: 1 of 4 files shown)


stable/node-problem-detector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: node-problem-detector
-version: 2.4.0
```
nyambati, line 3: Please update this to 2.4.1 as this is not breaking change

nyambati (changes requested):

stable/node-problem-detector/Chart.yaml
aslafy-z: Done

aslafy-z:
@nyambati I updated the docs to match the new chart version and rebased over master.

max-rocket-internet (approved):
Thanks @aslafy-z 💚

PR #789, opened 2026-04-15
[stable/node-local-dns] Make liveness probe configurable

paddy-bedraggled:
## Overview
This PR makes the `node-local-dns` chart liveness probe configurable instead of hardcoding the probe request to `/health` on `config.healthPort` with fixed timing values. The change is backward-compatible: existing installs keep the same effective probe behavior unless the new `livenessProbe` values are explicitly set.

## Context & Rationale
In our network, we want this service to listen only locally on the node for NodeLocal DNS traffic and not rely on externally reachable probe behavior. In particular, when the health listener is bound to a local node address, the liveness probe needs to be able to target that same host explicitly, for example via `livenessProbe.httpGet.host`.

Without this, the chart only allows configuring the probe port, which is not sufficient for environments where the health endpoint should remain local to the node and not be exposed more broadly just to satisfy liveness checks.

### Probe templating
- **`stable/node-local-dns/templates/daemonset.yaml`**: Replaced the hardcoded liveness probe with values-driven fields.
- Added support for:
  - `livenessProbe.httpGet.host`
  - `livenessProbe.httpGet.path`
  - `livenessProbe.httpGet.port`
  - `livenessProbe.httpGet.scheme`
  - `livenessProbe.initialDelaySeconds`
  - `livenessProbe.timeoutSeconds`
  - `livenessProbe.periodSeconds`
  - `livenessProbe.successThreshold`
  - `livenessProbe.failureThreshold`
- Kept `config.healthPort` as the fallback for the probe port so existing releases do not need to change values.

### Chart values
- **`stable/node-local-dns/values.yaml`**: Added a new `livenessProbe` configuration block with defaults matching the previous chart behavior.
- This allows users to set a local probe target such as:
  ```yaml
  livenessProbe:
    httpGet:
      host: 169.254.20.10
      port: 8080
  ```

### Documentation and versioning
- **`stable/node-local-dns/README.md`**: Documented the new `livenessProbe` values.
- **`stable/node-local-dns/Chart.yaml`**: Bumped the chart version from `2.7.0` to `2.7.1`.

## Verification
- [x] `helm lint stable/node-local-dns`
- [x] Rendered the chart with defaults using `helm template test stable/node-local-dns`
- [x] Verified the rendered default probe remains effectively unchanged:
  - path `/health`
  - port `8080` via `config.healthPort`
  - `initialDelaySeconds: 60`
  - `timeoutSeconds: 5`
- [x] Rendered the chart with `--set livenessProbe.httpGet.host=169.254.20.10 --set livenessProbe.httpGet.port=8080`
- [x] Confirmed the manifest emits `httpGet.host: "169.254.20.10"` in the daemonset probe

## Impact
- **Behavioral impact**: No change for existing users unless they opt into the new `livenessProbe` settings.
- **Compatibility**: Backward-compatible due to fallback to `config.healthPort`.
- **Reviewer focus**: Main review area is the daemonset probe templating and whether the added configuration surface is the right chart-level API.

## Task Tracking
- No existing upstream issue linked.

### DIFF
--- a/stable/node-local-dns/Chart.yaml
+++ b/stable/node-local-dns/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: node-local-dns
-version: 2.7.0
+version: 2.7.1
 appVersion: 1.26.7
 maintainers:
   - name: gabrieladt
--- a/stable/node-local-dns/README.md
+++ b/stable/node-local-dns/README.md
@@ -1,6 +1,6 @@
 # node-local-dns
 
-![Version: 2.7.0](https://img.shields.io/badge/Version-2.7.0-informational?style=flat-square) ![AppVersion: 1.26.7](https://img.shields.io/badge/AppVersion-1.26.7-informational?style=flat-square)
+![Version: 2.7.1](https://img.shields.io/badge/Version-2.7.1-informational?style=flat-square) ![AppVersion: 1.26.7](https://img.shields.io/badge/AppVersion-1.26.7-informational?style=flat-square)
 
 A chart to install node-local-dns.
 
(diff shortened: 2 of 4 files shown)


max-rocket-internet (approved):
Looks good but CI is failing

PR #792, opened 2026-04-21
[stable/cluster-overprovisioner] adding security defaults

Ro-Fo:
## Description

**[stable/cluster-overprovisioner] Harden security defaults and fix SecurityContext rendering**

This PR introduces security hardening for the `cluster-overprovisioner` chart by fixing a template rendering bug and setting least-privilege defaults out of the box.

### Bug fix: SecurityContext template rendering (`deployments.yaml`)

- **Container `securityContext`**: `toYaml ... | nindent 12` was placed inline after the key (`securityContext: {{ ... }}`), producing invalid YAML block mapping. Moved the value to a new line.
- **Pod `securityContext`**: The block was always rendered even when the value was an empty map (`{}`), resulting in `securityContext:\n  {}`. Wrapped in `{{- with ... }}` so it is only emitted when a value is set.

### Security defaults (`values.yaml`)

Hardened defaults are now set for the `pause` container, which requires no filesystem writes, no Linux capabilities, and no Kubernetes API access:

**`containerSecurityContext`**
- `allowPrivilegeEscalation: false`
- `readOnlyRootFilesystem: true`
- `runAsNonRoot: true` / `runAsUser: 65534` (nobody)
- `capabilities.drop: [ALL]`
- `seccompProfile.type: RuntimeDefault`

**`podSecurityContext`**
- `runAsNonRoot: true` / `runAsUser/Group: 65534`
- `fsGroup: 65534`
- `seccompProfile.type: RuntimeDefault`

**`serviceAccount.automountServiceAccountToken: false`**  
The `pause` pods make no Kubernetes API calls, so mounting a service account token is unnecessary and increases attack surface.

### Documentation (`README.md`)

Added a **Security** section explaining each hardened setting and the rationale behind it, including a note on how to override individual values if needed.

## Checklist

- [x] Title of the PR starts with chart name (e.g. `[stable/cluster-overprovisioner]`)

### DIFF
--- a/ci/README.md.gotmpl
+++ b/ci/README.md.gotmpl
@@ -43,6 +43,26 @@ helm install my-release oci://ghcr.io/deliveryhero/helm-charts/{{ template "char
 
 {{ template "chart.requirementsSection" . }}
 
+{{ if eq .Name "cluster-overprovisioner" -}}
+## Security
+
+The chart ships with hardened security defaults for both the pod and container level, following least-privilege principles.
+
+**Container security context** (`containerSecurityContext`):
+
+- `allowPrivilegeEscalation: false` — prevents the process from gaining additional privileges at runtime
+- `readOnlyRootFilesystem: true` — the container filesystem is mounted read-only
+- `runAsNonRoot: true` / `runAsUser: 65534` — container runs as the unprivileged `nobody` user
+- `capabilities.drop: [ALL]` — all Linux capabilities are dropped
+- `seccompProfile.type: RuntimeDefault` — syscall filtering via the container runtime's default seccomp profile
+
+**Pod security context** (`podSecurityContext`):
+
+- `runAsNonRoot: true` / `runAsUser: 65534` — enforced at pod level as an additional guard
+- `fsGroup: 65534` — volume ownership is set to `nobody`
+- `seccompProfile.type: RuntimeDefault` — seccomp applied at the pod level
+
+{{ end -}}
 {{ template "chart.valuesSection" . }}
 
 {{ template "chart.maintainersSection" . }}
--- a/stable/cluster-overprovisioner/README.md
+++ b/stable/cluster-overprovisioner/README.md
@@ -48,6 +48,24 @@ helm install my-release oci://ghcr.io/deliveryhero/helm-charts/cluster-overprovi
 * <https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler>
 * <https://github.com/kubernetes/kubernetes/tree/master/build/pause>
 
+## Security
+
+The chart ships with hardened security defaults for both the pod and container level, following least-privilege principles.
+
+**Container security context** (`containerSecurityContext`):
+
+- `allowPrivilegeEscalation: false` — prevents the process from gaining additional privileges at runtime
+- `readOnlyRootFilesystem: true` — the container filesystem is mounted read-only
+- `runAsNonRoot: true` / `runAsUser: 65534` — container runs as the unprivileged `nobody` user
+- `capabilities.drop: [ALL]` — all Linux capabilities are dropped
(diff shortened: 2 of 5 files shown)


stable/cluster-overprovisioner/README.md
```diff
@@ -48,6 +48,24 @@ helm install my-release oci://ghcr.io/deliveryhero/helm-charts/cluster-overprovi
...
+- `runAsNonRoot: true` / `runAsUser: 65534` — container runs as the unprivileged `nobody` user
+- `capabilities.drop: [ALL]` — all Linux capabilities are dropped
+- `seccompProfile.type: RuntimeDefault` — syscall filtering via the container runtime's default seccomp profile
+
+**Pod security context** (`podSecurityContext`):
+
+- `runAsNonRoot: true` / `runAsUser: 65534` — enforced at pod level as an additional guard
+- `fsGroup: 65534` — volume ownership is set to `nobody`
+- `seccompProfile.type: RuntimeDefault` — seccomp applied at the pod level
+
```
max-rocket-internet, line 68: You can remove all this section by just including correctly formatted comments in `stable/cluster-overprovisioner/values.yaml` as it will be autogenerated and automatically inserted into README.md.

I'll suggest changes to show you how...

stable/cluster-overprovisioner/values.yaml
```diff
@@ -1,11 +1,24 @@
 # containerSecurityContext -- Container security context object
-containerSecurityContext: {}
-# allowPrivilegeEscalation: false
+containerSecurityContext:
+  allowPrivilegeEscalation: false
```
max-rocket-internet, line 4: Like this:

```suggestion
  # containerSecurityContext.allowPrivilegeEscalation -- prevents the process from gaining additional privileges at runtime
  allowPrivilegeEscalation: false
```

max-rocket-internet: It's better, no?

max-rocket-internet:
@Ro-Fo you just need to regenerate the docs and push the changes

Ro-Fo:
@max-rocket-internet i think now its correct, thx

ci/README.md.gotmpl
```diff
@@ -43,6 +43,26 @@ helm install my-release oci://ghcr.io/deliveryhero/helm-charts/{{ template "char
...
+- `readOnlyRootFilesystem: true` — the container filesystem is mounted read-only
+- `runAsNonRoot: true` / `runAsUser: 65534` — container runs as the unprivileged `nobody` user
+- `capabilities.drop: [ALL]` — all Linux capabilities are dropped
+- `seccompProfile.type: RuntimeDefault` — syscall filtering via the container runtime's default seccomp profile
+
+**Pod security context** (`podSecurityContext`):
+
+- `runAsNonRoot: true` / `runAsUser: 65534` — enforced at pod level as an additional guard
+- `fsGroup: 65534` — volume ownership is set to `nobody`
+- `seccompProfile.type: RuntimeDefault` — seccomp applied at the pod level
```
max-rocket-internet, line 63: None of this is needed because the chart values are autogenerated, with the defaults and description. You just need to populate `stable/cluster-overprovisioner/values.yaml` correctly.

ci/README.md.gotmpl
```diff
@@ -43,6 +43,26 @@ helm install my-release oci://ghcr.io/deliveryhero/helm-charts/{{ template "char
...
+
+{{ end -}}
```
max-rocket-internet, line 65: I don't think we want to start putting customisations for each chart into the main template.

```suggestion
```

max-rocket-internet (changes requested):

ci/README.md.gotmpl
max-rocket-internet: Do you need help with this?

PR #795, opened 2026-05-12
[stable/node-local-dns]: add configurable container lifecycle hooks

omyani:
## Description

Currently, the node-local-dns chart does not allow users to define lifecycle hooks. This is a critical limitation for clusters using autoscalers like Karpenter or Cluster Autoscaler that frequently perform node consolidation.

The Problem:
When a node is drained, the kubelet sends a SIGTERM to all pods on the node. node-local-dns (a DaemonSet) often terminates immediately. However, application pods on the same node that are still performing a graceful shutdown (draining connections) may attempt final DNS lookups (e.g., to authentication services or logging endpoints). Since the local DNS cache is already dead, these apps surface ConnectTimeout or Temporary failure in name resolution errors, leading to 500s during deployment/consolidation events.

The Solution:
By adding support for lifecycle hooks, users can inject a preStop sleep. This forces the node-local-dns pod to wait (e.g., sleep 30) before terminating, ensuring it remains available to serve the last remaining application pods on the node.

## Checklist

### DIFF
--- a/stable/node-local-dns/Chart.yaml
+++ b/stable/node-local-dns/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: node-local-dns
-version: 2.7.0
+version: 2.8.0
 appVersion: 1.26.7
 maintainers:
   - name: gabrieladt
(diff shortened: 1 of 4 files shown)


max-rocket-internet (approved):
Nice, thanks @omyani 💚

PR #800, opened 2026-06-06
[stable/node-local-dns]: Correct logic wrapping securityContext

### LINKED ISSUE #797: node-local-dns chart gates securityContext with wrong logic
I was looking to use your node-local-dns Helm chart as this seems to be the best one out there however I found a bug.

The DaemonSet template incorrectly wraps the `securityContext` block with `{{ if .Values.config.setupIptables }}` so if you disable iptables as we do for Cilium it drops the security context settings as well.

bodgit:
## Description

This fixes the logic for the `securityContext` value so it's not gated by the iptables condition.

Fixes #797

## Checklist

### DIFF
--- a/stable/node-local-dns/Chart.yaml
+++ b/stable/node-local-dns/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: node-local-dns
-version: 2.9.0
+version: 2.9.1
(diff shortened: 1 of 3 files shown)


max-rocket-internet (approved):
Cool, thanks @bodgit 🩵

PR #802, opened 2026-06-17
[stable/node-local-dns]: add preDelete hook

bnagym:
## Description

The `node-local-dns` chart has no ordering control on `helm uninstall`. Helm deletes regular resources in its built-in order and does not wait for the DaemonSet pods to terminate before removing the resources those pods depend on at runtime — the ConfigMap and the `-upstream` Service.

When those dependencies are removed first, a pod that restarts or reschedules during the uninstall breaks: with the ConfigMap gone it cannot mount the volume and is stuck in `ContainerCreating`, and with the `-upstream` Service gone node-cache enters `CrashLoopBackOff`. The DaemonSet's deletion then blocks on these stuck pods and the release fails. The only way to recover was to force-delete the DaemonSet, which is not something that should be relied on as standard uninstall procedure.

This adds an optional `pre-delete` hook (default disabled) that gracefully drains the DaemonSet and waits for its pods to terminate before Helm removes its dependencies, so the pods shut down cleanly.

## Changes

- Add `hooks.preDelete` to `values.yaml` (default `enabled: false`, backwards-compatible)
- Add `templates/hooks/pre-delete/` with a dedicated ServiceAccount, Role, RoleBinding and drain Job, gated behind `hooks.preDelete.enabled`
- The Job gracefully deletes the DaemonSet (no `--force`) and waits for its pods before Helm proceeds; on timeout it fails the uninstall rather than leaving stuck pods
- Regenerate README
- Bump chart version to 2.10.0

## Checklist

### DIFF
--- a/stable/node-local-dns/Chart.yaml
+++ b/stable/node-local-dns/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: node-local-dns
-version: 2.9.1
+version: 2.10.0
 appVersion: 1.26.7
 maintainers:
   - name: gabrieladt
--- a/stable/node-local-dns/README.md
+++ b/stable/node-local-dns/README.md
@@ -1,6 +1,6 @@
 # node-local-dns
 
-![Version: 2.9.1](https://img.shields.io/badge/Version-2.9.1-informational?style=flat-square) ![AppVersion: 1.26.7](https://img.shields.io/badge/AppVersion-1.26.7-informational?style=flat-square)
+![Version: 2.10.0](https://img.shields.io/badge/Version-2.10.0-informational?style=flat-square) ![AppVersion: 1.26.7](https://img.shields.io/badge/AppVersion-1.26.7-informational?style=flat-square)
 
 A chart to install node-local-dns.
 
@@ -23,7 +23,7 @@ helm install --generate-name oci://ghcr.io/deliveryhero/helm-charts/node-local-d
 To install a specific version of this chart:
 
 ```console
-helm install --generate-name oci://ghcr.io/deliveryhero/helm-charts/node-local-dns --version 2.9.1
+helm install --generate-name oci://ghcr.io/deliveryhero/helm-charts/node-local-dns --version 2.10.0
 ```
 
 To install the chart with the release name `my-release`:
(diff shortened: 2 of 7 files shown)


max-rocket-internet:
> In that state they cannot terminate cleanly, so node-cache's iptables teardown never runs and node-local DNS keeps routing lookups to a dead local resolver.

Are you sure about this? Do you have a link to a related issue like this? What is specifically "node-cache's iptables"? I find it hard to believe k8s is not cleaning up IP tables rules just because pods are crashing when being removed.

bnagym:
> > In that state they cannot terminate cleanly, so node-cache's iptables teardown never runs and node-local DNS keeps routing lookups to a dead local resolver.
> 
> Are you sure about this? Do you have a link to a related issue like this? What is specifically "node-cache's iptables"? I find it hard to believe k8s is not cleaning up IP tables rules just because pods are crashing when being removed.

You're right, updated the description. Thanks for bringing this

What I can confirm is the actual problem: once Helm deletes the ConfigMap/-upstream Service before the DaemonSet pods are gone, the pods get stuck (ContainerCreating / CrashLoopBackOff) and the uninstall blocks on them and fails.

max-rocket-internet:
Can you paste the full terminal output of when you run `helm uninstall RELEASE_NAME`? Does it just hang?

bnagym:
> Can you paste the full terminal output of when you run `helm uninstall RELEASE_NAME`? Does it just hang?

Thank you for your fast responses!

I actually tested this exact uninstall on our 10+ node sandbox right before running it on Prod and it completed there, so I can't reproduce the failure on sandbox to grab clean output. Below what we observed on Prod.

Setup: uninstall runs via Terraform (`helm_release`), chart: 2.3.0, EKS: 1.33, 100+ nodes. When the uninstall via Terraform failed, we ran `helm uninstall` directly with different flags.

Sequence:
1. TF apply failed: `Error: failed to delete release: nodelocal-dns`. Helm had already deleted the ConfigMap, nodelocal-dns-upstream Service, and ServiceAccount, but the DaemonSet was stuck.
2. Re-ran (wrongly assuming flaky apply) — same error.
3. Ran `helm uninstall nodelocal-dns -n kube-system` directly — didn't complete (no captured output).
4. Tried `--no-hooks --ignore-not-found` and other flags — none completed it.
5. Daemonset pods wedged in **ContainerCreating/CrashLoopBackOff**: `configmap "nodelocal-dns" not found`, and after we restored the ConfigMap, `plugin/forward: not an IP address or file: "<nil>"` (Service was gone too).
6. Tried a normal `kubectl delete ds` — didn't succeed. Force-deleting it (`--force --grace-period=0`) finally let Helm finalize the release.

I'm not certain node-count is the real differentiator. FYI during the uninstall, nodes were actively scaling up/down. 

It could be a deletion-ordering problem: Helm removed the Daemonset's startup/runtime dependencies (ConfigMap  +  Service) before its pods were gone, so when the pods were recreated they couldn't start — and couldn't cleanly terminate either. The pre-delete hook guarantees the Daemonset pods are fully gone first before Helm touches the dependent resources.

max-rocket-internet:
OK I see. You have many different parts involved there, terraform, helm provider and helm itself. I'm not convinced a fix for this issue should be in the helm chart itself. Especially as a `Job` + RBAC etc 🙂

bnagym:
> OK I see. You have many different parts involved there, terraform, helm provider and helm itself. I'm not convinced a fix for this issue should be in the helm chart itself. Especially as a `Job` + RBAC etc 🙂

For me the uncomfortable part is that even if there is some issue on the provider side, why `helm uninstall` didn't pass?

What solutions you suggest? I want to ensure that Daemonset will be gracefully deleted first before dependent resources.

For now I have only workaround to patch the daemonset by appending non-existing label by using NodeSelector -> Pods are safely deleted from the nodes and "scheduled" to a non-existing ones:

```shell
kubectl -n kube-system patch daemonset nodelocal-dns --type=merge -p '{
  "spec": {
    "template": {"spec": {"nodeSelector": {"non-existing/label": "true"}}}
  }
}'
```

max-rocket-internet:
> For me the uncomfortable part is that even if there is some issue on the provider side, why helm uninstall didn't pass?

I get it but I'm just not convinced this a problem that should be solved in the chart itself? Maybe create an issue in the helm repo?

> I actually tested this exact uninstall on our 10+ node sandbox right before running it on Prod and it completed there, so I can't reproduce the failure on sandbox to grab clean output.

You seem to have a lot of uncertainties. Perhaps spend more time to debug it?

Or alternatively apply your workaround in terraform, like `null_resource` to run a command that does the clean up with `kubectl`, just like the `Job` in this PR 🙂

PR #803, opened 2026-06-19
[stable/k8s-event-logger] update to 2.7 release

al-cheb:
Upgrade k8s-event-logger from version 2.6 to 2.7.

## Checklist

- [x] Title of the PR starts with chart name (e.g. `[stable/mychartname]`)
- [x] I have read the [contribution instructions](https://github.com/deliveryhero/helm-charts#opening-a-pr), bumped chart version and regenerated the docs
- [ ] Github actions are passing

PR #804, opened 2026-06-22
[stable/no-chart] Update all github actions

max-rocket-internet:
We are seeing this error in the checkout actions on [this PR](https://github.com/deliveryhero/helm-charts/pull/803):

```
Run actions/checkout@v7.0.0
  with:
    repository: deliveryhero/helm-charts
    token: ***
    ssh-strict: true
    ssh-user: git
    persist-credentials: true
    clean: true
    sparse-checkout-cone-mode: true
    fetch-depth: 1
    fetch-tags: false
    show-progress: true
    lfs: false
    submodules: false
    set-safe-directory: true
    allow-unsafe-pr-checkout: false
/usr/bin/docker exec  04b576a44ce4a6871737aa1e5367819751dbeecb1cfa1d283c02f1d3a19ac1d4 sh -c "cat /etc/*release | grep ^ID"
Error relocating /__e/node24_alpine/bin/node: pthread_getname_np: symbol not found
```

So I'll just try a low-effort update of all actions first...

### DIFF
--- a/.github/workflows/chart-testing.yml
+++ b/.github/workflows/chart-testing.yml
@@ -6,22 +6,22 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v4.1.7
+        uses: actions/checkout@v7.0.0
         with:
           fetch-depth: 0
 
       - name: Set up Helm
-        uses: Azure/setup-helm@v3.3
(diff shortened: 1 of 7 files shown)


max-rocket-internet:
The only mention I can find of this error for github actions is here: https://github.com/actions/checkout/issues/2246

PR #806, opened 2026-07-02
[stable/node-local-dns] feat: add serve_stale support to cache plugin

skyalbrigonia:
Adds `serve_stale` option to the node-local-dns cache plugin configuration.


max-rocket-internet (approved):
Nice, thanks @skyalbrigonia 🩵

PR #809, opened 2026-07-25
[stable/kube-bench]: Security hardening + update to v0.15.6

danielqb:
## Summary

Security hardening of the kube-bench chart + update to latest kube-bench v0.15.6 (CIS Kubernetes Benchmark 1.10, Kubernetes 1.28-1.31).

### Version bump
- **appVersion**: 0.8.0 → v0.15.6
- **chart version**: 0.1.17 → 0.1.18

### Security fixes (7 Datree misconfigurations)
1. **Resources**: Added CPU/memory requests (100m/128Mi) and limits (256Mi)
2. **startingDeadlineSeconds**: 300s for CronJob
3. **securityContext**: runAsNonRoot, readOnlyRootFS, drop ALL capabilities
4. **serviceAccount.create**: true by default + RBAC (ClusterRole/ClusterRoleBinding)
5. **Probes**: Optional liveness/readiness probes
6. **Labels**: Added app.kubernetes.io/component: compliance
7. **Volumes**: Complete mounts for etcd, kube-scheduler, controller-manager, CNI, systemd

## Testing
```bash
helm lint stable/kube-bench/
helm template test stable/kube-bench/ --set serviceAccount.create=true --set provider=generic -n test-ns
```


stable/kube-bench/templates/_helpers.tpl
```diff
@@ -42,6 +42,7 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
+app.kubernetes.io/component: compliance
```
max-rocket-internet, line 45: This is not a standard label. If people want this, or other random labels added, should that not be done with an `extraLabels` value or similar?

stable/kube-bench/values.yaml
```diff
@@ -7,34 +7,42 @@ provider: eks
...
+    drop:
+      - ALL
 
 extraLabels: {}
 podLabels: {}
-resources: {}
+resources:
+  requests:
+    cpu: 100m
+    memory: 128Mi
```
max-rocket-internet, line 35: Best practice is to set these the same 🙂

```suggestion
    memory: 256Mi
```

